PR #418, opened 2015-01-12
Fixing a try finally error

tavinathanson:
Nit, but: when connection wasn't created, the finally's `.close()` errored (variable wasn't declared)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/418)

### DIFF
--- a/common/helpers.py
+++ b/common/helpers.py
@@ -14,13 +14,10 @@ def tables(database, *table_names):
             ...
       Where vcfs and genotypes are tables in the provided db.
     """
-    try:
-        connection = database.engine.connect()
+    with database.engine.connect() as connection:
         metadata = sqlalchemy.MetaData(bind=connection)
         metadata.reflect()
         yield tuple([connection] + [metadata.tables[t] for t in table_names])
-    finally:
-        connection.close()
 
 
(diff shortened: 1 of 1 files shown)


common/helpers.py
```diff
@@ -14,13 +14,10 @@ def tables(database, *table_names):
             ...
       Where vcfs and genotypes are tables in the provided db.
     """
-    try:
-        connection = database.engine.connect()
+    with database.engine.connect() as connection:
```
ihodes: Hmm, I wrote this code, but the `finally` (even if the connection were in scope) looks like it would be a problem? If the connection is closed, then it can't be used in the function using `tables`, right? Does this still work?

ihodes:
This LGTM

common/helpers.py
tavinathanson: Does this answer your question, or did you mean something else?

https://docs.python.org/2/library/contextlib.html
"At the point where the generator yields, the block nested in the with statement is executed. The generator is then resumed after the block is exited"

tavinathanson:
Cool, I'll merge once I fix Travis

PR #419, opened 2015-01-12
Some CSS fixes

### LINKED ISSUE #388: Let the /runs table get a bit wider
There's plenty of room on the page to avoid wrapping these three columns:

screen shot 2014-12-15 at 12 18 03 pm

(seen on cycledash.demeter)

### LINKED ISSUE #413: export as svg: what button to press
This is probably just confusion on my end, but when I press the print icon ("export publication quality") I get a menu that let's me pick the format, but I can't find the button to press to get the export to happen. Any hints?

image

danvk:
This tightens over-broad rules which were affecting `<table>`s they weren't meant to (see #413).
It also makes the Runs table more careful about what wraps when space is limited.

Before:
screen shot 2015-01-12 at 5 18 03 pm

After:
screen shot 2015-01-12 at 5 18 23 pm

I made whatever CSS changes were necessary to avoid dpxdt diffs (i.e. to make this a visual no-op). We can tweak the styles if we like.

Fixes #388
Fixes #413 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/419)

### DIFF
--- a/cycledash/static/css/runs.css
+++ b/cycledash/static/css/runs.css
@@ -43,6 +43,10 @@ td span.run-id {
   font-size: 0.92em;
 }
 
+.runs .run-id, .runs .date {
+  white-space: nowrap;
+}
+
 /* drag and drop file upload */
 .receiving-drag {
   background: rgb(247, 252, 253);
--- a/cycledash/static/css/examine.css
+++ b/cycledash/static/css/examine.css
@@ -20,35 +20,41 @@ body {
 .vcf-table {
   padding-top: 20px;
 }
-th {
+.vcf-table th {
   font-size: 19px;
   position: relative;
   white-space: nowrap;
 }
-th.arrow, td.arrow {
+.vcf-table th.arrow,
+.vcf-table td.arrow {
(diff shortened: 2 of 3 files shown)


cycledash/static/css/runs.css
```diff
@@ -43,6 +43,10 @@ td span.run-id {
   font-size: 0.92em;
 }
 
+.runs .run-id, .runs .date {
```
ihodes: Not really set on either way, but a simpler alternative is adding a `nowrap` (or, to be more semantic, `grouped-text`?) class, instead of a class for each column we don't want to wrap?

ihodes:
LGTM % 1 thought

cycledash/static/css/runs.css
danvk: I prefer it with `.run-id`, `.date` and friends. That way the display stuff stays entirely in the CSS file and the classes in the HTML file have clear non-visual meanings.

ihodes: Makes sense, works for me!

PR #423, opened 2015-01-15
Fix a comment bubble alignment issue

tavinathanson:
Fixing this:

image

It (the comment bubble getting separated from the comment count) happens when caller names get too long. I decided not to include a test for the case where the comment bubble overflows (caller name too long), but I could be convinced otherwise.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/423)


ihodes:
LGTM, no test seems reasonable (we have many other things we should probably test before worrying about this to much ;)

Issue #407, opened 2014-12-17
Order contig & position correctly


ihodes:
We're ALMOST there: 

![](https://s3.amazonaws.com/f.cl.ly/items/253C0O2p3w0W2A2O2e06/Screen%20Shot%202014-12-17%20at%2012.36.03%20PM.png)

We sort by contig length, and then lexicographically. We need to do a little better…

arahuja:
Not sure if this the same issue or a new one, but looks like positions are not sorted correctly now either:

![image](https://cloud.githubusercontent.com/assets/455755/5631278/54e23fde-9593-11e4-88ea-5f004d7f1031.png)

From [this run](http://hammerlab-dev3.hpc.mssm.edu/runs/190/examine?query=%7B%22range%22%3A%7B%22start%22%3Anull%2C%22end%22%3Anull%2C%22contig%22%3Anull%7D%2C%22filters%22%3A%5B%5D%2C%22sortBy%22%3A%5B%7B%22order%22%3A%22asc%22%2C%22columnName%22%3A%22contig%22%7D%2C%7B%22order%22%3A%22asc%22%2C%22columnName%22%3A%22position%22%7D%5D%2C%22page%22%3A1%2C%22limit%22%3A250%7D)

ihodes:
FWIW, while I work on the issue + translate this into SQLAlchemy, the SQL to sort contig/position nicely looks like: 

```SQL
order by COALESCE(SUBSTRING(contig FROM '^\d+')::INTEGER, 1000), contig, position;
```

ihodes:
This regressed somehow and passed tests: http://cycledash.demeter.hpc.mssm.edu/runs/153/examine

Need to fix and add better regression testing.

ihodes:
This appears to only be a problem when contigs are prefixed with chr.

PR #433, opened 2015-01-23
Fix ordering by position

ihodes:
Position was no longer being sorted by correctly, due to casting its type
to a string.

As noted by @arahuja in issue #407

This does not fix the issue, though; that requires a more complex sort
by contig (in the works).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/433)


ihodes:
This is failing because of the pylint requirement (don't know why) which is fixed in #434. Good to merge?

danvk:
LGTM

danvk:
Does this (or the previous PR) affect any screenshots?

ihodes:
The other one ended up changing a screenshot; I've added it here.

PR #434, opened 2015-01-23
Contigs are sorted correctly in genotypes.py

### LINKED ISSUE #407: Order contig & position correctly

ihodes:
Sort contigs first by number, then length, then lexicographically

Fixes #407

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/434)

### DIFF
--- a/cycledash/genotypes.py
+++ b/cycledash/genotypes.py
@@ -4,7 +4,9 @@ import copy
 
 from sqlalchemy import (select, func, types, cast, join, outerjoin, asc, desc,
                         and_, Integer, Float, String)
+from sqlalchemy.sql import text
 from sqlalchemy.sql.expression import label, column
+from sqlalchemy.sql.functions import coalesce
 import vcf as pyvcf
 from plone.memoize import forever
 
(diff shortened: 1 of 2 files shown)


cycledash/genotypes.py
```diff
@@ -108,9 +110,15 @@ def get(vcf_id, query, with_stats=True, truth_vcf_id=None):
         q = _add_filters(q, g, query.get('filters'))
         q = _add_orderings(q, g, query.get('sortBy'))
         q = _add_paging(q, g, query.get('limit'), query.get('page'))
+<<<<<<< Updated upstream
```
danvk: merge markers!

cycledash/genotypes.py
```diff
@@ -108,9 +110,15 @@ def get(vcf_id, query, with_stats=True, truth_vcf_id=None):
...
         q = q.order_by(asc(func.length(g.c.contig)),
                        asc(g.c.contig),
                        asc(g.c.position))
+=======
```
danvk: did anything actually change here? `git diff` might just be getting confused by the new blank line above.

danvk: nm, was commenting on an old diff

cycledash/genotypes.py
```diff
@@ -108,9 +110,9 @@ def get(vcf_id, query, with_stats=True, truth_vcf_id=None):
         q = _add_filters(q, g, query.get('filters'))
         q = _add_orderings(q, g, query.get('sortBy'))
         q = _add_paging(q, g, query.get('limit'), query.get('page'))
-        q = q.order_by(asc(func.length(g.c.contig)),
-                       asc(g.c.contig),
-                       asc(g.c.position))
+
+        q = _add_ordering(q, g, 'String', 'contig', 'asc')
```
danvk: `column_type` doesn't get used in this case, and the `contig` column certainly isn't sorted like a string. I'd just set it to something like `None` or `"n/a"` or `"(special)"` to avoid confusion.

danvk:
LGTM % 1.

We talked a bit last month about using the dpxdt DB to run simple python tests. A regression test for this issue would be a great place to start.

cycledash/genotypes.py
ihodes: Yeah, I went back and forth on this one and settles on 'String' because the user of _add_ordering needn't care about how contig is handled specially. Instead, they should just pass the parameters like they would for any other ordering that needs adding.

ihodes:
This is covered by a dpxdt SS pushed in #433

PR #437, opened 2015-01-26
Align X in CQL bad query

### LINKED ISSUE #296: CQL X alignment isn't always correct
On Chrome 39.0.2171.65, OS X:

image

When I set `.query-status.bad:before ` to `left: 0px`, it centers properly

ihodes:
Can't repro screenshot (could be a font issue?)

fixes #296

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/437)


danvk:
add a screenshot? LGTM otherwise.

tavinathanson:
Oops, missed this one, :+1:

ihodes:
@danvk There is a screenshot—not sure what you mean?

danvk:
Shouldn't this PR show a change in the screenshot then?

ihodes:
Do you mean the dpxdt screenshots? None of those show bad queries, so this wouldn't change any of them.

danvk:
Yeah, I'm talking about dpxdt. That's why I suggested adding a screenshot -- it's a UI state that's currently untracked.

ihodes:
Ah, I thought you were referring to a screenshot, not a test! 

I'm a little averse to adding a screenshot test for a little thing like this given that each tests adds about 5+ seconds to my tests (and probably longer for @tavinathanson). Is the marginal benefit we gain from tracking states like these worth the repeated pain?

danvk:
We should solve that by figuring out how to make the tests run faster, not by avoiding writing tests!

ihodes:
I agree 100%.

ihodes:
Finally did this; good to go? The test looks wonky due to the font we're using… this is something we should probably work out. At some pt, these should be icons anyway.

tavinathanson:
@isaac if the bug is the alignment of the X, and the alignment looks wonky in the test, that sounds to me like an unhelpful test?

tests/pdifftests/test.yaml
```diff
@@ -78,6 +78,18 @@ tests:
...
         }, 250);
 
+  - name: examine-bad-query
+    description: Examine page showing a poorly formed query
+    url: http://localhost:5001/runs/1/examine
+    config:
+      <<: *std-config
+      injectJs: |
+        setTimeout(function() {
+          q = $('.query-status').first();
```
danvk: FYI the `.first()` isn't necessary (but it doesn't hurt when there's only one matched element).

danvk:
LG, thanks for adding the screenshot!

ihodes:
@tavinathanson yeah, there's just not a way to make it work otherwise, unless we add icons. Problem is with PhantomJS + our custom font.

PR #438, opened 2015-01-27
Standardize on JSON for VCF data structure serialization

### LINKED ISSUE #406: Python arrays are stored in Postgres as strings: should be CSVs or PG arrays
E.g. extant_columns is serialized as a string __repr__ of a Python array, which makes manipulating it difficult. Similarly with VCF arrays which get inserted into the DB. We should probably store them as comma-separated lists.

ihodes:
When VCFs are serialized to CSV and then inserted into postgres, arrays
are now serialized to JSON.

This also adds a test for serializing back to VCF, and cleans up some code.

fixes #406

This is a **WIP**, still; needs further testing and potential migration code for the production CD. I'm also not entirely sold on serializing to JSON; will report back if I have further thoughts on this.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/438)


ihodes:
@danvk @tavinathanson and I talked about this some. Two issues come up when serializing VCF values to JSON:

1) We would just be serializing arrays, so we'll conditionally be ser/deserializing things depending on the derived type of the field (e.g. some VCF's XX field is an array, some are just a single integer). [We won't be serializing e.g. string values, as all that gives us is strings wrapped in quotation marks in the DB.]
2) More importantly, all we get from the arrays is an extra bracket (and potentially quotation marks around the strings, if it's an array of strings) in the DB; this makes running regexps and other queries over these values much harder. 

As such, I think storing things in the DB as they're stored in the VCF (e.g. 1,2,3 in a VCF => 1,2,3 in the DB) is the way to go. (At least until we enforce some stricter schema on our data; then we can have properly typed DB columns and store arrays, ints, etc, instead of just TEXT types.)

If this makes sense to both of you, I'll pull out the tests in this PR as a separate PR, and close this one.

hammer:
If you're serializing arrays, why not use PG's array data type?

ihodes:
That would be ideal, but right now we need all our columns to be of type TEXT, as we cannot enforce any schema on this VCF data. This is something I'd love to change, but right now VCFs have the same fields with conflicting types, so I just went with TEXT for now. 

cf. 
> At least until we enforce some stricter schema on our data; then we can have properly typed DB columns and store arrays, ints, etc, instead of just TEXT types.

PR #457, opened 2015-02-17
Reorganize the project's toplevel files

ihodes:
I wanted to see how cleaning up the toplevel would feel. Seems to feel good. Let me know if I should proceed with this @dan @tavinathanson 

(This is a partial commit; docs and scripts need to be modified before
this is cohesive.)

WIP: I'd like feedback on this before rewriting the docs/scripts to make this change work.

https://github.com/hammerlab/cycledash/tree/reorg-top-level to see how it looks

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/457)


danvk:
I'm fine with moving most files out of the top level dir, except the ones for which this is a canonical location. So:

- `LICENSE` -- my sense is that if there isn't a top-level LICENSE file, people (or code!) may assume there is no license.
- `CONTRIBUTING.md` -- I think this needs to be top-level for github to show the link when you create a new issue/PR
- `Makefile` -- maybe just turn this into a script instead? I think it's broken at the moment.
- `gulpfile.js` -- It seems fairly standard to run `gulp` from the top-level dir to iterate on JS. What would this look like with it in `/scripts`?

tavinathanson:
If we're doing some thinking about folder organization, I think it'd be nice to have that thinking written up in our style guide and used for similar projects. 

In terms of these specific changes, some of it feels a little non-standard? (And I generally vote for going the standard route when it comes to organization.) Documentation, for example, seems to usually be at the toplevel?

ihodes:
Good points. 

@danvk 

1) Re license, there's still a toplevel "license folder"; that would seem sufficient for any honest user.
2) CONTRIBUTING.md, ah, probably so unfortunately. Good catch.
3) Makefile — yeah, good point.
4) gulpfile — also a good point, and a shame. I dislike all these toplevel files, but I guess it makes sense…

@tavinathanson 

Good idea re: styleguide, though different stacks result in a vastly different project structure.

The README is still toplevel, and can link to other docs/wiki pages etc. That seems enough to me, though it seems we'll need CONTRIBUTING.md to be toplevel anyway.

tavinathanson:
Right, so style for semi-large Flasky Reacty projects? Still good to put it out there, even if it needs tweaking, IMO.

ihodes:
:+1: I agree

danvk:
Closing this for the time being, feel free to re-open if you want to revisit.

PR #472, opened 2015-02-26
Organize runs page by project

### LINKED ISSUE #467: Make the runs page a data-browsing page
After talking with @arahuja about this, the /runs page should be a way to browse the datasets, projects, and runs we have in a multifaceted way. E.g. we should be able to group runs by dataset, group runs by project, sort + filter them. Maybe CQL on top of the runs page with aggregate/GROUP BY would be sufficient. 

Otherwise, a stopgap measure would be to make a /projects page & a /datasets page.

ihodes:
The projects are sorted so that the project with the most recent run is
at the top, and the runs are sorted by most recent at the top within
each project.

This is the first step to reorganizing the runs page, and one that
requires no changes to the database.

fixes #467 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/472)

### DIFF
--- a/cycledash/static/js/runs/components/RunsPage.js
+++ b/cycledash/static/js/runs/components/RunsPage.js
@@ -24,8 +24,7 @@ var RunsPage = React.createClass({
     completions: React.PropTypes.object.isRequired
   },
   getInitialState: function() {
-    return {selectedRunId: null,
-            projectFilter: null,
+    return {projectFilter: null,
             showForm: false,
             draggingOver: false};
   },
@@ -37,12 +36,6 @@ var RunsPage = React.createClass({
       return noFilter || matches;
     });
   },
-  handleClickRun: function(runId) {
-    return () => {
-      var selectedRunId = this.state.selectedRunId === runId ? null : runId;
-      this.setState({selectedRunId});
-    };
-  },
   handleProjectFilter: function(evt) {
     var val = evt.target.value;
     this.setState({projectFilter: val === NO_FILTER ? null : val});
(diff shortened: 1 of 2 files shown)


ihodes:
I need to add the dpxdt screenshots, but I'm getting some weird errors I'll fix later. Figured this could go to review before then.

danvk:
@arahuja you cool w/ this UI?

cycledash/static/js/runs/components/RunsPage.js
```diff
@@ -64,9 +57,71 @@ var RunsPage = React.createClass({
...
     });
     var runs = this.filteredRuns();
-    var rows = runs.map(run => {
-      var rows = [<RunRow run={run} key={run.id}
-                          handleClick={this.handleClickRun(run.id)} />];
+    var projectRuns = _.chain(runs)
+        .groupBy('project_name')
+        .map((runs, projectName) => [projectName, runs])
+        .value();
+    var projectTables = _.chain(projectRuns).sortBy(function([projectName, runs]) {
```
danvk: merge this into the `chain`?

cycledash/static/js/runs/components/RunsPage.js
```diff
@@ -64,9 +57,71 @@ var RunsPage = React.createClass({
...
+      return -runs[0].id; // to sort by the descending ID
+    }).map(function([projectName, runs]) {
```
danvk: is there a bug out for the fat-arrow/destructuring clash? it's a shame.

danvk:
Looks fine code-wise

cycledash/static/js/runs/components/RunsPage.js
ihodes: Good idea! Done.

arahuja:
Yea this looks good, its the same datafields, but grouped into projects right?

ihodes:
Yeah, simple change for now. Will keep the PRs coming to get to the point we were talking about.

cycledash/static/js/runs/components/RunsPage.js
ihodes: I haven't actually looked… may be worth doing. It's definitely a shame.

PR #476, opened 2015-02-27
Add NT field to schema so that field can be parsed out of VCFs

### LINKED ISSUE #471: NT column in info field is not parsed out
`INFO` field looks like 
```
QSS_ref NT=hom;QSS=55;QSS_NT=55;SGT=GG->GT;SOMATIC;TQSS=1;TQSS_NT=1
```

NT is not showing as a column, but QSS and on are.

ihodes:
fixes #471

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/476)

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -127,6 +127,7 @@ CREATE TABLE genotypes (
        "info:SGT" TEXT, -- Most likely somatic genotype excluding normal noise states
        "info:TQSS" TEXT, -- [INTEGER] Data tier used to compute QSS
        "info:TQSS_NT" TEXT, -- [INTEGER] Data tier used to compute QSS_NT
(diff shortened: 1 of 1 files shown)


schema.sql
```diff
@@ -61,6 +61,7 @@ CREATE TABLE genotypes (
        "info:MQ" TEXT, -- RMS mapping quality
        "info:MQ0" TEXT, -- [INTEGER] number of mapQ == 0 reads covering this record
        "info:NS" TEXT, -- [INTEGER] number of samples with data
+       "info:NT" TEXT, -- [INTEGER] number of samples with data
```
arahuja: comment line seems wrong?

danvk:
Do you need to update the dpxdt database dump?

ihodes:
@arahuja & @danvk  good catch, will change

ihodes:
@danvk back to you; don't need to edit the dpxdt data dump it turns out.

PR #478, opened 2015-02-27
Link to IGV from the Examine page

### LINKED ISSUE #466: Link to IGV
It should be possible to add a "View in IGV" link to the examine page:
http://www.broadinstitute.org/igv/ControlIGV

cc @arahuja

danvk:
Fixes #466 

Here's what it looks like:
screen shot 2015-02-27 at 3 30 39 pm

This isn't entirely trivial for the user—it involves some one-time setup. See docs at https://github.com/hammerlab/cycledash/wiki/IGV-Integration

I wouldn't say I'm happy about threading `igvLink` through five components—@ihodes let me know if you have any better idea.

cc @arahuja @ryan-williams @timodonnell 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/478)


cycledash/static/css/examine.css
```diff
@@ -150,14 +150,21 @@ body {
...
   line-height: 1.6em;
+}
+.vcf-table .dalliance-open {
+  background: url(/static/img/magnifying-glass-noun_907.svg) no-repeat 2px 6px;
+  cursor: pointer;
   background-size: 15px 15px;
 }
+.vcf-table .igv-links {
+  background: url(/static/img/igv32x32.jpg) no-repeat 2px 3px;
+  background-size: 16px 16px;
```
ihodes: Why not 15x15 like the other one?

cycledash/static/js/examine/components/VCFTable.js
```diff
@@ -471,17 +480,28 @@ var VCFComment = React.createClass({
...
+
     return (
       <div className='comment-header'>
         <a className='dalliance-open'
-           onClick={() => {this.props.handleOpenViewer(this.props.record);}}>
+           onClick={() => {this.props.handleOpenViewer(r);}}>
           Open Pileup Viewer
         </a>
+        <span className='igv-links'>
+          IGV: <a href={loadIGVLink}>Load at Locus</a>&nbsp;
```
ihodes: No need to do this in this PR unless you'd like + it's trivial, but could be nice to disable the "load at locus" once it's been clicked once, as after the first time you load, you'll just need the "jump" button.

cycledash/static/js/examine/utils.js
```diff
@@ -19,4 +19,27 @@ function juxt(fns) {
...
+  var nameFilePairs = [
+      ['Run', run.uri],
+      ['Normal', run.normal_bam_uri], 
+      ['Tumor', run.tumor_bam_uri]
+  ].filter(x => x[1]);
+
+  var fileParam = nameFilePairs.map(x => fileUrl(x[1])).join(','),
+      nameParam = nameFilePairs.map(x => x[0]).join(',');
+
+  return `http://localhost:60151/load?user=cycledash&genome=hg19` +
```
ihodes: Factor out magic string?

tests/js/ExaminePage-comments-test.js
```diff
@@ -48,14 +48,15 @@ describe('ExaminePage Comments', function() {
       created_at: '',
       uri: '/tests/js/data/snv.vcf'
     };
+  var igvHttpfsUrl = 'http://example.com/';
```
ihodes: Why do we need example.com?

ihodes:
Looks great, this is super exciting. Couple minor comments & good to merge. I think this needs dpxdt screenshots?

ihodes:
+ I agree, this props threading is pretty absurd. I don't have a better solution, though, but it bears some thinking about. Definitely the most tedious thing about changing a React app.

cycledash/static/css/examine.css
danvk: I think 16x16 is a more standard icon size? The `6px` offset vs. the `3px` offset means they don't line up precisely anyway. The different shapes of the two icons make that more about how it looks anyway.

cycledash/static/js/examine/components/VCFTable.js
danvk: Or change it to `IGV: Jump to Locus (reload)` to shift the emphasis. Filed #479

tests/js/ExaminePage-comments-test.js
danvk: Done

cycledash/static/js/examine/utils.js
danvk: Done

danvk:
I've updated the screenshots. Looks like we forgot to update them after your "projects" change, too.

The "3 months ago" text is going to be a problem with screenshots going forward. We could either hard-code the date or black that out before taking a screenshot.

PR #492, opened 2015-03-02
Qualified import, because Python is weird.

ihodes:
For some reason, cycledash.genotypes and cycledash.comments were referring to functions, not the imported modules. 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/492)

### DIFF
--- a/cycledash/__init__.py
+++ b/cycledash/__init__.py
@@ -32,13 +32,15 @@ def _configure_logging(app):
 def _configure_application(app):
     app.config.from_object('config')
 
+
 def _configure_templates(app):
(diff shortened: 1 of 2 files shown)


danvk:
It's because of [this line](https://github.com/hammerlab/cycledash/blob/master/cycledash/__init__.py#L44):

```python
# cycledash/__init__.py
from cycledash.views import *
```

can we delete that and avoid this in the future?

ihodes:
Nice catch. Fixed.

danvk:
I pulled this locally & confirmed that it fixes the issue for me. Please merge on green.

PR #500, opened 2015-03-02
Add tests for relational_vcf.py

ihodes:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/500)

### DIFF
--- a/tests/python/test_relational_vcfs.py
+++ b/tests/python/test_relational_vcfs.py
@@ -1,0 +1,89 @@
+import nose
+
+import StringIO
+
+import vcf, csv
+from common.relational_vcf import vcf_to_csv, genotypes_to_records
+
+
+def test_vcf_to_csv():
+    vcfreader = vcf.Reader(open('tests/data/snv.vcf'))
+    columns = ['contig', 'position', 'sample:DP', 'info:VAF', 'info:DP', 'something']
+    filename = vcf_to_csv(vcfreader, columns, None,
+                          default_values={'something': 'Something!'})
+
+    with open(filename) as f:
+        rows = list(csv.reader(f))
+
+    # first two are from the same sample
+    assert rows[0] == ['20', '61795', '44', '', '81', 'Something!']
+    assert rows[1] == ['20', '61795', '37', '', '81', 'Something!']
+    # this is the last sample
+    assert rows[-1] == ['20', '75254', '40', '', '74', 'Something!']
+    assert len(rows) == 20  # 10 records, 2 samples each = 20 rows
+
+
+def test_csv_to_vcf():
+    # Here we just write the VCF to a CSV like we did above, and then read it
+    # and compare it to the original VCF to make sure we haven't lost any data
+    # in the conversation.
+    vcfreader = vcf.Reader(open('tests/data/one-sample.vcf'))
+    columns = ['contig', 'position', 'id', 'reference', 'alternates',
+               'quality', 'filters', 'info:DP', 'info:SS',
+               'info:SSC', 'info:GPV', 'info:SPV', 'info:SOMATIC',
+               'sample:GT','sample:GQ','sample:DP','sample:RD',
+               'sample:AD','sample:FREQ','sample:DP4',
+               'sample_name']
+    filename = vcf_to_csv(vcfreader, columns, None)
+
+    # Now we convert the rows in this CSV into dict (converting the position as
+    # well), which is what it would look like select from Postgres.
+    with open(filename) as f:
+        rows = list(csv.reader(f))
+        relations = []
+        for row in rows:
+            relation = {
+                col: val if val != '' else None
+                for col, val in zip(columns, row)
+            }
+            relation['position'] = int(relation['position'])
+            relations.append(relation)
+
+    text = open('tests/data/one-sample.vcf').readlines()
+    header = (line for line in text if line.startswith('#'))
+    template = vcf.Reader(header)
+
+    # Here we convert those dicts to vcf.Records, for later writing.
+    records = genotypes_to_records(relations, template, columns)
+
+    # Now we write the output of those to a string buffer...
+    vcf_sb = StringIO.StringIO()
+    writer = vcf.Writer(vcf_sb, template)
+    for record in records:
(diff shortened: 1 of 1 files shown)


tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
+import nose
+
+import tempfile
+
+
+
+def test_vcf_to_csv():
+    import vcf, csv
+    from common.relational_vcf import vcf_to_csv
```
danvk: what happens if you move these imports to the top?

tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
...
+    vcfreader = vcf.Reader(open('tests/data/snv.vcf'))
+    columns = ['contig', 'position', 'sample:DP', 'info:VAF', 'info:DP', 'something']
+    filename = vcf_to_csv(vcfreader, columns, None,
+                          default_values={'something': 'Something!'})
+
+    with open(filename) as fd:
```
danvk: I always think of file descriptors as numbers. Do they usually refer to streams in Python?

tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
...
+        rows = list(csv.reader(fd))
+
+    # first two are from the same sample
+    assert rows[0] == ['20', '61795', '44', '', '81', 'Something!']
+    assert rows[1] == ['20', '61795', '37', '', '81', 'Something!']
+    # this is the last sample
+    assert rows[-1] == ['20', '75254', '40', '', '74', 'Something!']
+    assert len(rows) == 20 # 10 records, 2 samples each = 20 rows
```
danvk: two spaces between end of line and `#`

tests/python/test_relational_vcfs.py
ihodes: Nothing; don't recall why I put them there. Maybe to keep code specific to a tests in the test only? Regardless… moved them to the top.

tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
...
+
+    text = open('tests/data/one-sample.vcf').readlines()
+    header = (l for l in text if l.startswith('#'))
+    template = vcf.Reader(header)
+
+    # Here we convert those dicts to vcf.Records, for later writing.
+    records = genotypes_to_records(relations, template, columns)
+
+    # Now we write the output of those to a file...
+    with tempfile.NamedTemporaryFile(mode='w', delete=False) as fd:
```
danvk: `vcf.Writer` takes a stream, so you should be able to use `StringIO` and avoid the file system.

tests/python/test_relational_vcfs.py
ihodes: Not sure I follow—they refer to "file objects" [according to the docs](https://docs.python.org/2/library/functions.html#open), but I've often seen `fd` (file descriptor) used in `with` statements.

tests/python/test_relational_vcfs.py
ihodes: :+1:

tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # first two are from the same sample
+    assert rows[0] == ['20', '61795', '44', '', '81', 'Something!']
+    assert rows[1] == ['20', '61795', '37', '', '81', 'Something!']
+    # this is the last sample
+    assert rows[-1] == ['20', '75254', '40', '', '74', 'Something!']
+    assert len(rows) == 20 # 10 records, 2 samples each = 20 rows
+
+
+def test_csv_to_vcf():
```
danvk: add a comment indicating that this is testing that `vcf_to_csv` and `csv_to_vcf` are inverses—it's actually testing more than just `csv_to_vcf` (it would be doing just that if you checked in a CSV file to tests/data, for example, rather than generating it)

tests/python/test_relational_vcfs.py
danvk: That page doesn't use the word "descriptor" or `fd`. Wikipedia seems to agree that they're ints: [File descriptor](http://en.wikipedia.org/wiki/File_descriptor).

I'd be if the var was just called `f`. I'm bringing this up because I began reading this code assuming that `fd` was a number and got confused.

tests/python/test_relational_vcfs.py
```diff
@@ -0,0 +1,90 @@
...
+        for row in rows:
+            relation = {
+                col: val if val != '' else None
+                for col, val in zip(columns, row)
+            }
+            relation['position'] = int(relation['position'])
+            relations.append(relation)
+
+    text = open('tests/data/one-sample.vcf').readlines()
+    header = (l for l in text if l.startswith('#'))
```
danvk: single letter vars called `l` are a pet peeve—in a lot of fonts they look identical to `I` or `1`. Just make it `line` or `x`.

tests/python/test_relational_vcfs.py
ihodes: I can change it to `f`, no problem. Didn't mean to indicate that it's an integer (or a stream?).

tests/python/test_relational_vcfs.py
ihodes: Done

tests/python/test_relational_vcfs.py
danvk: It is a stream, isn't it?

ihodes: It [implements the iterator protocol](https://docs.python.org/2/library/stdtypes.html#bltin-file-objects), so, sure!

tests/python/test_relational_vcfs.py
ihodes: Done

tests/python/test_relational_vcfs.py
ihodes: Done

danvk:
LGTM, feel free to merge on green

PR #497, opened 2015-03-02
Put CQL queries in the URL, not JSON

danvk:
This results in much cleaner URLs (see `test.yaml` for before/after).

This also fixes a few scary bugs:

- `get_all_comments` always returned `None`
- The behavior of `NULL` and `NOT NULL` was reversed.
- `_add_filter` constructed all possible operators, then threw all but one away. This wouldn't work for null checks, since `x > NULL` is invalid SQL.
- `toString()` was broken for queries involving null checks

One thing to note: this parses the query and then sets the URL parameter by re-serializing it to CQL. A simpler approach might be to store the raw CQL query in the URL box, but this would involve threading that through the `RecordStore`. I'm open to suggestions.

Fixes #495 
Fixes #304 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/497)

PR #501, opened 2015-03-03
Link to comments

### LINKED ISSUE #455: Link directly to individual comments
The /runs page shows recent comments and links to the run in which they were made. But it doesn't link to the specific comment. It would be nice if there were a way to go straight to the variant in question.

An easy way to do this would be by pre-filling the CQL query to be something like "chr3:12345" to only show that variant. It would be better to scroll down to the relevant variant, but this would require some backend changes, e.g. to only load variants 100,000-100,500 if the comment was on the 100,100th variant.

danvk:
(diffbasing off #497 since Travis is being slow.)

Fixes #455 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/501)


ihodes:
LGTM

PR #507, opened 2015-03-03
Only show true/false/positive stats with validation data

### LINKED ISSUE #376: Don't show true/false/positive/negative when there's no validation data
These numbers don't make sense without a validation run. The "Showing all 1,234 variants" line is still useful, though.

danvk:
Fixes #376 

Apparently this used to work but hasn't in some time due to a [triple-equal `null` check][1] on a non-existent property.

[1]: https://github.com/hammerlab/cycledash/blob/d90f2398a8cb5f72d9e9785e4b8ed85a1204697c/cycledash/static/js/examine/components/StatsSummary.js#L15

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/507)


ihodes:
Add a dpxdt test for the case where there is a validation set? LGTM otherwise!

danvk:
There are some tests with validation sets (e.g. [validated-run.png][1]), but they don't show up here because not a single pixel changed.

[1]: https://github.com/hammerlab/cycledash/blob/master/tests/pdifftests/validated-run.png

PR #510, opened 2015-03-04
Annotate datasets

### LINKED ISSUE #475: Allow datasets to be annotated
E.g. in Confluence we have a table of datasets with things like "notes", "date of resection", "location of sample", etc. We should support that in CycleDash, on the /runs page.

### LINKED ISSUE #493: Write python tests
See https://github.com/hammerlab/cycledash/pull/492 — we should at least have enough coverage that this would break Travis tests.

### LINKED ISSUE #524: Random autocompletes
Why are these options shown in the autocomplete?
image

Even more, it changes as I go from "stand" to "standard" but none actually match (in a way that I can tell)

### LINKED ISSUE #532: Gene Annotation failing due to change in pyensembl
```
[2015-03-10 16:16:52,988: ERROR/MainProcess] Task workers.gene_annotator.annotate[94ec72fa-8dac-4516-8519-5b337fe3429f] raised unexpected: ValueError('Ensembl annotation data is not currently installed for release 75. Run "pyensembl install 75" or call EnsemblRelease(75).install()',)
```

ihodes:
Sorry for the massive PR; this seemed like a better way to do it than in pieces, so that I could evolve the API with the UI. I'll try to make it as painless as possible with a high-level overview here (The TODO list is also a decent list of the changes that happened). Comments in commits should also be helpful. 

The views.py is now almost completely routing code, all requests and responses are dealt with in `cycledash.genotypes`, `cycledash.runs`, `cycledash.bams`, `cycledash.projects`.

`cycledash.projects` and `cycledash.bams` are new APIs for projects and bams in the database. It's worth checking out schema.sql early on to see how things relate to one another. BAMs belong to projects, and VCFs also belong to projects. Projects are the top-level object, and can correspond to patients or other projects (e.g. DREAM, DREAM-Synth3, or Testing).

`tests/python/test_api.py` is also a useful file for seeing how the new APIs work (cf. API.md and /about as well).

Most the other changes are on the front-end, adding forms for projects and BAMs, and allowing the /runs view to also display BAMs.

I'll be hosting a local version at mach-3.local:5000, so you can check out the changes there as well.

This PR generated some issues as well: #519 #520, #530.

Migration SQL

``` sql
CREATE TABLE projects (
       id BIGSERIAL PRIMARY KEY,
       name TEXT UNIQUE NOT NULL,
       notes TEXT
);

CREATE TABLE bams (
       id BIGSERIAL PRIMARY KEY,
       project_id BIGINT REFERENCES projects NOT NULL,
       name TEXT,
       normal BOOLEAN DEFAULT FALSE, -- true if the bam is of normal tissue
       notes TEXT,
       tissues TEXT,
       resection_date TEXT,
       uri TEXT NOT NULL
);


alter table vcfs add column tumor_bam_id bigint references bams;
alter table vcfs add column normal_bam_id bigint references bams;
alter table vcfs add column project_id bigint references projects;
alter table vcfs add column normal boolean default false;


update vcfs set project_name = 'Default' where project_name is null;
insert into projects (name)
  select distinct project_name from vcfs where project_name is not null;

insert into bams (uri, name, project_id, normal)
  select distinct V.normal_bam_uri, V.dataset_name, P.id, true
  from vcfs V, projects P
  where V.project_name = P.name
      and v.normal_bam_uri is not null;

insert into bams (uri, name, project_id)
  select distinct V.tumor_bam_uri, V.dataset_name, P.id
  from vcfs V, projects P
  where V.project_name = P.name
      and v.tumor_bam_uri is not null;

update vcfs set normal_bam_id = bams.id
  from bams
  where vcfs.dataset_name = bams.name
    and bams.uri = vcfs.normal_bam_uri
    and bams.normal = true;

update vcfs set tumor_bam_id = bams.id
  from bams
  where vcfs.dataset_name = bams.name
    and bams.uri = vcfs.tumor_bam_uri
    and bams.normal = false;

update vcfs set project_id = projects.id
  from projects
  where projects.name = vcfs.project_name;


alter table vcfs drop column project_name;
alter table vcfs drop column tumor_bam_uri;
alter table vcfs drop column normal_bam_uri;
alter table vcfs drop column datase_name;
```
#### TODO:
- [x] BAMs & Projects API
- [x] New /runs UI for BAMs and Projects
- [x] Testing story for API
- [x] Migration SQL for production/local Postgres
- [x] Allow `project_name` or `project_id` (similarly with `bam_{uri,id}`)
- [x] Add autocomplete for BAMs back
- [x] Update documentation (/about, README, etc.)
- [x] Add API tests for errors/validation of projects/bams/etc.
- [x] Show required fields in Forms
- [x] dpxdt tests
- [x] Migration SQL

Fixes #524, fixes #475, fixes #493, fixes #532

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/510)


danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)

Reviewed files:
- .gitignore @ r1
- API.md @ r1
- common/helpers.py @ r1
- config.py @ r1
- cycledash/bams.py @ r1
- cycledash/comments.py @ r1
- cycledash/genotypes.py @ r2
- cycledash/helpers.py @ r1
- cycledash/projects.py @ r1
- cycledash/runs.py @ r2
- cycledash/static/css/runs.css @ r1
- cycledash/static/css/style.css @ r5
- cycledash/static/js/examine/utils.js @ r3
- cycledash/static/js/runs/components/Forms.js @ r5
- cycledash/static/js/runs/components/RunsPage.js @ r5
- cycledash/static/js/runs/runs.js @ r1
- cycledash/templates/about.html @ r1
- cycledash/templates/runs.html @ r1
- cycledash/validations.py @ r1
- cycledash/views.py @ r1
- DEVELOP.md @ r1
- README.md @ r1
- requirements.txt @ r1
- schema.sql @ r1
- scripts/lint.sh @ r1
- scripts/travis-setup.sh @ r1
- scripts/travis-test.sh @ r1
- tests/create-test-db.sh @ r1
- tests/data.sql @ r1
- tests/ENV.sh @ r1
- tests/js/examine-utils-test.js @ r4
- tests/pdifftests/create-test-db.sh @ r1
- tests/pdifftests/data.sql @ r1
- tests/pdifftests/start-cycledash.sh @ r1
- tests/pdifftests/test.yaml @ r1
- tests/python/test_api.py @ r1
- tests/start-cycledash.sh @ r1
- workers/gene_annotator.py @ r6
- workers/genotype_extractor.py @ r1
- workers/runner.py @ r6

---

Mostly looks good! Thanks for writing the tests.

My high level questions are:

- How does specifying a validation VCF work in this new world?
- Why move Flask view code into `runs.py` and friends?

---

<sup>**[.gitignore, line 14 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk3z3sAuBuIVHuYc1zU)** ([raw file](https://github.com/hammerlab/cycledash/blob/2396c09b1bcac8f26edd8d2ac1151fcee7181b87/.gitignore#L14)):</sup>
This should be under `cycledash/static/lib/dalliance` now. Try deleting this path from your repo & see if it comes back.

---

<sup>**[API.md, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45xcESlli0VnwfYTX)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L1)):</sup>
This feels a lot like a type definitions file. Are there standard tools people use to describe their REST APIs in computer-readable formats? Alex just showed me http://rest.ensembl.org/, for example, which looks generated.

---

<sup>**[API.md, line 5 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45Tv9sM7Sq-wKpaUF)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L5)):</sup>
\` instead of &lt;tt&gt; since this is markdown?

---

<sup>**[API.md, line 27 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45eHuracNHnkTwF7u)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L27)):</sup>
what happens if it isn't?

---

<sup>**[API.md, line 39 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45lV24gZ-9_i9dq87)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L39)):</sup>
constant (sp)

---

<sup>**[common/helpers.py, line 61 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Hdeyt_oKq2l66S6a)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/common/helpers.py#L61)):</sup>
How about `findFirst`?

---

<sup>**[common/helpers.py, line 61 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4HlQunIbmacvOVsew)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/common/helpers.py#L61)):</sup>
`lst` → `iterable`?

---

<sup>**[common/helpers.py, line 61 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4ORB_KmNRF_m8Vzbg)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/common/helpers.py#L61)):</sup>
Given how you use this, it might make more sense to have the API be:

```python
foo = find(collection, key1=value1, key2=value2)
```

rather than passing in `lambda`s.

---

<sup>**[cycledash/bams.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4IA3A-Y9lG99Bh5tJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L1)):</sup>
Add top-level comment for this module.

---

<sup>**[cycledash/bams.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4KPJSewGDJWOgmV24)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L1)):</sup>
I don't follow the rationale behind moving Flask response logic (i.e. `error_response`, `jsonify`, `redirect`) into this file and out of `views.py`. This file would make sense (and be easier to test) if it was just focused on interacting with the DB, rather than both interacting with the DB and generating Flask responses.

---

<sup>**[cycledash/bams.py, line 17 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4LzQc4zTJxeMkP9sg)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L17)):</sup>
use `get_where`?

---

<sup>**[cycledash/bams.py, line 23 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4IsrLHQv_Or0fJgO9)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L23)):</sup>
`helpers. error_response`?

---

<sup>**[cycledash/bams.py, line 34 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JnwyC09LLKtxBySY)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L34)):</sup>
Add:

```
args:
    request: cycledash.validations.CreateBamSchema
```

or equivalent

---

<sup>**[cycledash/bams.py, line 46 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JSFQ-61EsND5pDpJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L46)):</sup>
why do this? seems like it'd make the API more error-prone for anyone consuming the error vs. just always returning an array.

---

<sup>**[cycledash/bams.py, line 62 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4J5pYfx2GquTQUXsN)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L62)):</sup>
Add `# 201: Created`

---

<sup>**[cycledash/bams.py, line 64 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JhQs-U4C0ngzKqdQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L64)):</sup>
else error?

---

<sup>**[cycledash/bams.py, line 68 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JvV1oKfWL_QFKPrL)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L68)):</sup>
document expected type of `request`

---

<sup>**[cycledash/genotypes.py, line 87 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4LdU4eM3qFmPKkodT)** ([raw file](https://github.com/hammerlab/cycledash/blob/1676d1cccc8e9ad593f24bfd6cf2ba9f44163948/cycledash/genotypes.py#L87)):</sup>
was this dead code?

---

<sup>**[cycledash/helpers.py, line 67 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4M8yvGbEhTaQm2RdK)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/helpers.py#L67)):</sup>
Worth mention that **kwargs are meant to be strings, rather than `table.c.column_name`, which is the syntax that SQLAlchemy normally uses.

---

<sup>**[cycledash/projects.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Ml8AqNc8CCIZwNKU)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/projects.py#L1)):</sup>
There's a ton of duplication between this and `bams.py`. You could probably factor out the generation of a basic CRUD REST API given a table.

---

<sup>**[cycledash/projects.py, line 14 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4MZ8qDJcgBcfPvd43)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/projects.py#L14)):</sup>
This function never returns `None`.

---

<sup>**[cycledash/runs.py, line 21 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4PmQf441R_-R-F7Ap)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/runs.py#L21)):</sup>
How about returning `vcfs, projects` from this function and having a separate function to generate the project tree?

---

<sup>**[cycledash/runs.py, line 141 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4QksAEF5VQDitN_6p)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/runs.py#L141)):</sup>
`"""After calling this, run['(bam_type)_bam_id'] will be set to a valid BAM ID."""`

---

<sup>**[cycledash/runs.py, line 142 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4R3YzxXR1hKazD3C8)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/runs.py#L142)):</sup>
spaces around `+` operator

---

<sup>**[cycledash/static/js/examine/utils.js, line 38 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4RTQZTb3X7QKGMRS-)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/examine/utils.js#L38)):</sup>
How could `run.normal_bam` be something without a `uri`?

---

<sup>**[cycledash/static/js/runs/components/Forms.js, line 70 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4S1U2YS-fMs4bm1e_)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/Forms.js#L70)):</sup>
What's `role=form`?

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 3 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SEAhve6w0dk-SxBy)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L3)):</sup>
capital F `Forms`

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 14 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SN8lsAGloZWXFvOq)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L14)):</sup>
Document the type of `projects` and `comments` more precisely, either here or somewhere else that you reference from here.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 22 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SZm6c46pR6_wG8Tr)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L22)):</sup>
`setDisplayProjectForm` (It's surprising that a method called `displayProjectForm` would actually hide the project form.)

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 30 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Sku50hVgq-dWx1qt)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L30)):</sup>
Indent continuation lines 4 spaces.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 46 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TgD3IYL-DjZdb6ug)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L46)):</sup>
I'm a little confused by the combination of this line & the previous one -- `project.vcfs[0]` is both a number and an Object?

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 54 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TvlsPqg50FS7e_ZJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L54)):</sup>
arrow instead of `.bind(this)`?

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 55 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L55)):</sup>
In light of my discovery about re-rendering subcomponents due to re-creating these functions on every `render`, you may want to split `displayProjectForm` into two methods, `showProjectForm` and `hideProjectForm`.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 91 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4UI-C2QvQ7ovOI-Ej)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L91)):</sup>
long line -- put each of these onto its own line.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 237 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4UqiXVJVpgXuUoYgv)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L237)):</sup>
This might belong in [statics](http://facebook.github.io/react/docs/component-specs.html#statics).

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 348 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4V-BbMp54kOELcN8Z)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L348)):</sup>
I suspect `.string` is incorrect :)

In any case, please document the type more precisely.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 386 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VCj4DkTrkETUn3K1)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L386)):</sup>
kill

---

<sup>**[cycledash/templates/about.html, line 18 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VTmLlNUIiPL0xSFn)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/templates/about.html#L18)):</sup>
This should only be stored in one place—the two will get out of sync.

One idea would be to read `API.md` in the `/about` handler and format it. Or just say "See API.md" on this page.

---

<sup>**[cycledash/validations.py, line 70 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L70)):</sup>
Is it customary to suffix these functions with `Schema`? If I were writing type definitions, I'd call this `UpdateProjectRequest`. The `Schema` makes me think it's some kind of database schema.

---

<sup>**[cycledash/views.py, line 147 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wf9VuH4gpMfaS5TF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/views.py#L147)):</sup>
VCF Upload/Download

---

<sup>**[schema.sql, line 16 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wt4WzCISXLB7ka-r)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L16)):</sup>
enum?

---

<sup>**[schema.sql, line 17 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wz2nXtgr8WYDMAhp)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L17)):</sup>
add comment. what's a library?

---

<sup>**[schema.sql, line 19 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wv0b14WblExA7JTs)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L19)):</sup>
I can't parse this comment. What's a met?

---

<sup>**[tests/create-test-db.sh, line 2 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XGwhB3jAXAWfF8ye)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/create-test-db.sh#L2)):</sup>
you don't say!

(not a useful comment given the file name -- either remove or flesh it out)

---

<sup>**[tests/python/test_api.py, line 9 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L9)):</sup>
yeah, these wouldn't be necessary if `projects.py` weren't Flask-y.

---

<sup>**[tests/python/test_api.py, line 122 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4YY0WXiVzYwdeM5cA)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L122)):</sup>
this should be in its own test file (it's testing a different module than the other test fixture).

---

<sup>**[tests/python/test_api.py, line 186 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4YhaJzC8BroQhlwS7)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L186)):</sup>
holey money that's a lot of zeros!

---

<sup>**[workers/gene_annotator.py, line 32 \[r6\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Yrj4pO7hRmZftBHU)** ([raw file](https://github.com/hammerlab/cycledash/blob/2dea170d19e1b81cf6a7afa44f514969b6b552d2/workers/gene_annotator.py#L32)):</sup>
two spaces before `#`

---

<sup>**[workers/gene_annotator.py, line 32 \[r6\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4YuRLaWb9NKSGPEcE)** ([raw file](https://github.com/hammerlab/cycledash/blob/2dea170d19e1b81cf6a7afa44f514969b6b552d2/workers/gene_annotator.py#L32)):</sup>
make `77` an environment var?

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

(I reviewed r5, so it's possible my comments are out of date)

---

ihodes:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

> How does specifying a validation VCF work in this new world?

Via a dropdown on the /examine page (or a multiselect box). Cf. issue [#520](https://github.com/hammerlab/cycledash/issues/520).

> Why move Flask view code into `runs.py` and friends?

`comments.py` was already this way, and it save saves `views.py` from needing to know about all the failure cases of the other modules. This way, `views.py` is basically `routes.py` (and I can rename it now, or when the final few errantly large functions are moved out of it).

---

<sup>**[API.md, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45xcESlli0VnwfYTX)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L1)):</sup>
I agree that it'd be useful to use a generator from e.g. validation.py. I think that could be done in another PR/at another time. For now, this is a copy-paste of the HTML from /about, so it's quite trivial to keep in sync.

---

<sup>**[API.md, line 5 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45Tv9sM7Sq-wKpaUF)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L5)):</sup>
This is just a copy-paste from the HTML, which makes it trivial to keep in sync (the reason the headers are markdown is because I used a tool which transformed HTML -> MD... it didn't work too well.

---

<sup>**[API.md, line 27 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45eHuracNHnkTwF7u)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L27)):</sup>
You'll get a JSON error messages explaining that it isn't.

---

<sup>**[common/helpers.py, line 61 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Hdeyt_oKq2l66S6a)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/common/helpers.py#L61)):</sup>
`find` usually implies returning the first one (in underscore, for example); it's the short-circuiting version of `filter`

---

<sup>**[common/helpers.py, line 61 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4ORB_KmNRF_m8Vzbg)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/common/helpers.py#L61)):</sup>
That'd work too, but I'd rather keep the flexibility/keep it like other `find`s in other languages/libraries.

---

<sup>**[cycledash/bams.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4KPJSewGDJWOgmV24)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L1)):</sup>
There are a few reasons. It's worth noting that `comments.py` was already structured this way. The main move is to separate the routing in views.py from the response generation in the other modules. The other modules are primary concerned with catching errors and response generation. The extraction and inserting from and to the DB isn't something I'm interested in testing apart from how it works to generate the API. As you can see from the tests, testing the API is actually quite straightforward, and is the functionality I want to be testing. 

---

<sup>**[cycledash/bams.py, line 36 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JnwyC09LLKtxBySY)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L36)):</sup>
Removed argument instead. It's just a Flask request, not a Schema.

---

<sup>**[cycledash/bams.py, line 66 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4J5pYfx2GquTQUXsN)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L66)):</sup>
Seems a bit superfluous to document common HTTP codes?

---

<sup>**[cycledash/bams.py, line 72 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JvV1oKfWL_QFKPrL)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L72)):</sup>
Removed argument instead. It's just a Flask request.

---

<sup>**[cycledash/genotypes.py, line 86 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4LdU4eM3qFmPKkodT)** ([raw file](https://github.com/hammerlab/cycledash/blob/1676d1cccc8e9ad593f24bfd6cf2ba9f44163948/cycledash/genotypes.py#L86)):</sup>
Was what dead code? Reviewable is showing me this: http://cl.ly/image/0s2c0o1V1W2a

---

<sup>**[cycledash/projects.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Ml8AqNc8CCIZwNKU)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/projects.py#L1)):</sup>
I agree, and I was definitely tempted to, but I figured this would be a good start.

---

<sup>**[cycledash/static/js/examine/utils.js, line 38 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4RTQZTb3X7QKGMRS-)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/examine/utils.js#L38)):</sup>
This ensures we don't to try access the `uri` attribute on undefined, if the `run` doesn't have a `normal_bam`.

---

<sup>**[cycledash/static/js/runs/components/Forms.js, line 68 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4S1U2YS-fMs4bm1e_)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/Forms.js#L68)):</sup>
It's an accessibility thing, but doesn't make sense here give the lack of accessibility of the rest of our app. Removed.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 3 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SEAhve6w0dk-SxBy)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L3)):</sup>
Hmm, `Forms` -> `forms` is the way to do it here, as Forms.js is not a single component.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 22 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SZm6c46pR6_wG8Tr)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L22)):</sup>
I think it's less confusing when you see it being called: `displayProjectForm(false)` is pretty declarative. It also mirrors CSS's `display: none`. 

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 55 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L55)):</sup>
I don't think that'll cause any problems here; it seems much simpler this way, too.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 237 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4UqiXVJVpgXuUoYgv)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L237)):</sup>
It says there that statics is for static methods?

---

<sup>**[cycledash/templates/about.html, line 18 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VTmLlNUIiPL0xSFn)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/templates/about.html#L18)):</sup>
Cf. comments on API.md.

---

<sup>**[cycledash/validations.py, line 70 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L70)):</sup>
They *are* schemas, so naming them that way makes it clear what they are when used in other modules. 

I'm not sure why you'd suffix it with `request`?

---

<sup>**[schema.sql, line 16 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wt4WzCISXLB7ka-r)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L16)):</sup>
removed this column

---

<sup>**[schema.sql, line 17 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wz2nXtgr8WYDMAhp)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L17)):</sup>
removed this column

---

<sup>**[schema.sql, line 19 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wv0b14WblExA7JTs)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L19)):</sup>
metastasis 

removed this column anyway

---

<sup>**[tests/create-test-db.sh, line 2 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XGwhB3jAXAWfF8ye)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/create-test-db.sh#L2)):</sup>
But *[you](https://github.com/hammerlab/cycledash/blame/master/tests/pdifftests/create-test-db.sh)* did ;) 

Deleted it.

---

<sup>**[tests/python/test_api.py, line 9 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L9)):</sup>
I'd still want to test the JSON API, not solely that I can insert things into the database. 

---

<sup>**[workers/gene_annotator.py, line 31 \[r6\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Yrj4pO7hRmZftBHU)** ([raw file](https://github.com/hammerlab/cycledash/blob/2dea170d19e1b81cf6a7afa44f514969b6b552d2/workers/gene_annotator.py#L31)):</sup>
Already fixed in next commits

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

Looking good, mostly small stuff from here.

---

<sup>**[API.md, line 5 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk45Tv9sM7Sq-wKpaUF)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/API.md#L5)):</sup>
Nothing ever stays in sync unless you enforce it, no matter how easy it is. Feel free to file an issue if you don't want to do it in this PR, but we really shouldn't have duplicate documentation.

---

<sup>**[cycledash/bams.py, line 63 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4J5pYfx2GquTQUXsN)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L63)):</sup>
Common in what world? :) I'd never heard of it until I saw it here & looked it up.

---

<sup>**[cycledash/bams.py, line 65 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JhQs-U4C0ngzKqdQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L65)):</sup>
not resolved?

---

<sup>**[cycledash/genotypes.py, line 87 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4LdU4eM3qFmPKkodT)** ([raw file](https://github.com/hammerlab/cycledash/blob/1676d1cccc8e9ad593f24bfd6cf2ba9f44163948/cycledash/genotypes.py#L87)):</sup>
http://cl.ly/image/1W1I0x210g2h

---

<sup>**[cycledash/projects.py, line 1 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Ml8AqNc8CCIZwNKU)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/projects.py#L1)):</sup>
k, please file an issue for it

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 22 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SZm6c46pR6_wG8Tr)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L22)):</sup>
But CSS property names are nouns and method names are verbs.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 45 \[r10\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk5MTGb7yLBfIe4xb8l)** ([raw file](https://github.com/hammerlab/cycledash/blob/645fac107862c5551e3b6d8a441aeab2dce4c7de/cycledash/static/js/runs/components/RunsPage.js#L45)):</sup>
drop parens

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 46 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TgD3IYL-DjZdb6ug)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L46)):</sup>
I'm still confused.

```
> x = {id: 1}
Object {id: 1}
> -x
NaN
> -(undefined)
NaN
```

So won't `vcf` always be `NaN` here?

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 54 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TvlsPqg50FS7e_ZJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L54)):</sup>
not resolved?

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 55 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L55)):</sup>
Maybe we disagree about function/method names usually being verbs? I feel like this has come up before.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 237 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4UqiXVJVpgXuUoYgv)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L237)):</sup>
Yeah, it's unclear to me if it's for static methods or all static things. In any case, this is fine.

---

<sup>**[cycledash/validations.py, line 69 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L69)):</sup>
Maybe I'm misunderstanding how voluptuous is being used, but the `Schema` feels a bit like adding `Type` to the end of all your data types. I'd call this `UpdateProjectRequest` because it's a type which describes the shape of a request to update a project.

---

<sup>**[schema.sql, line 19 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Wv0b14WblExA7JTs)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/schema.sql#L19)):</sup>
Cool. I'm a fan of less structure.

---

<sup>**[tests/python/test_api.py, line 9 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L9)):</sup>
Actually, why do you need this? You could just use `self.app.post('/projects')` in your tests. I don't think you're gaining anything by avoiding two API calls in your tests.

---

<sup>**[tests/python/test_bams_api.py, line 41 \[r10\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk5Po-HdJA3nORh85eb)** ([raw file](https://github.com/hammerlab/cycledash/blob/645fac107862c5551e3b6d8a441aeab2dce4c7de/tests/python/test_bams_api.py#L41)):</sup>
this is overly elaborate—now that these fields aren't in the schema, setting `data` to something like `hello world` and then checking that it comes back out is enough. This isn't a test for `json.dumps` and `json.loads`, after all.

(same comment applies to `data` fields below)

---

<sup>**[workers/gene_annotator.py, line 32 \[r6\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4YuRLaWb9NKSGPEcE)** ([raw file](https://github.com/hammerlab/cycledash/blob/2dea170d19e1b81cf6a7afa44f514969b6b552d2/workers/gene_annotator.py#L32)):</sup>
My suggestion was to make this part of `ENV.sh`/`config.py`.

---

<sup>**[workers/gene_annotator.py, line 39 \[r10\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk5Qh4I880dJxSfd5Yz)** ([raw file](https://github.com/hammerlab/cycledash/blob/645fac107862c5551e3b6d8a441aeab2dce4c7de/workers/gene_annotator.py#L39)):</sup>
Put this below the `register_running_task` call—otherwise the worker will potentially run untracked for a long time as it downloads the Ensembl data set.

---

ihodes:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

Cool, hopefully good to go now!

---

<sup>**[cycledash/bams.py, line 63 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4J5pYfx2GquTQUXsN)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L63)):</sup>
I'd thought common in the world of HTTP API design/webapps; anytime you create a resource via POST, a 201 Created should be returned. I'll add a comment to all 201 codes!

---

<sup>**[cycledash/bams.py, line 65 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JhQs-U4C0ngzKqdQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L65)):</sup>
Ah I thought you were stating that it would error; this is actually the code you wrote in one of your recent commits, I didn't change it, just moved it. I've figured that if a client doesn't accept HTML or JSON content, an error is fine. Feel free to file an issue to address this all over CycleDash, it's probably good to just default to just JSON; I think it's outside the scope of this PR given that it's been the case for a while.

---

<sup>**[cycledash/genotypes.py, line 86 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4LdU4eM3qFmPKkodT)** ([raw file](https://github.com/hammerlab/cycledash/blob/1676d1cccc8e9ad593f24bfd6cf2ba9f44163948/cycledash/genotypes.py#L86)):</sup>
Ah, yes, dead code. Too bed Reviewable doesn't show me what you're commenting on… 

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 23 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4SZm6c46pR6_wG8Tr)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L23)):</sup>
"display" is also a verb. I'll change it since it's confusing, though!

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 47 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TgD3IYL-DjZdb6ug)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L47)):</sup>
Oops, the - is a typo, fixed!

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 56 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TvlsPqg50FS7e_ZJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L56)):</sup>
It doesn't work. function() is required for whatever reason. 

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 57 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L57)):</sup>
`display` is a verb; I'm not sure what the issue really is: your initial comment suggests that there's some inefficiency in using a closure here, and this subsequent comment is talking about something else? Is it really unclear what `displayProjectForm(false)` does? If so, I will change it, I'm just not sure if that's what you want? Is there a preferable verb-phrase for this method (other than making two redundant methods?)

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 243 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4UqiXVJVpgXuUoYgv)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L243)):</sup>
> The statics object allows you to define static methods that can be called on the component class.

Implies methods only, so I didn't try it/this seemed fine.

---

<sup>**[cycledash/validations.py, line 69 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L69)):</sup>
I'll change it to not have `Schema` at the end. This is another thing that's been in CD for months.

---

<sup>**[tests/python/test_api.py, line 0 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L0)):</sup>
I need this helper function so that I'm not writing a lot of duplicate code, otherwise I'd repeat those lines of code or the equivalent (and more verbose) `.post` code five times. There's no reason to insert a project into the DB with HTTP instead of directly via the database API. Is there a reason why you'd prefer using an additional API call, or why you think the function isn't necessary?

---

<sup>**[tests/python/test_bams_api.py, line 41 \[r10\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk5Po-HdJA3nORh85eb)** ([raw file](https://github.com/hammerlab/cycledash/blob/645fac107862c5551e3b6d8a441aeab2dce4c7de/tests/python/test_bams_api.py#L41)):</sup>
These fields are in the schema. I want to make sure the API works as advertised, so I'm testing these fields. Setting `data` to `hello world` would error and not accomplish what the test is trying to accomplish. What is elaborate here?  This seems pretty standard.

---

<sup>**[workers/gene_annotator.py, line 31 \[r6\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4YuRLaWb9NKSGPEcE)** ([raw file](https://github.com/hammerlab/cycledash/blob/2dea170d19e1b81cf6a7afa44f514969b6b552d2/workers/gene_annotator.py#L31)):</sup>
Sure. FWIW this has been this way for months, I've just fixed a bug in this worker in this PR.

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

Looks good!

---

<sup>**[cycledash/bams.py, line 62 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4J5pYfx2GquTQUXsN)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L62)):</sup>
It's possible it's completely common, just unfamiliar to me—I've always done APIs with [Stubby](http://www.quora.com/What-RPC-does-google-use-with-Protocol-Buffers).

---

<sup>**[cycledash/bams.py, line 65 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JhQs-U4C0ngzKqdQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L65)):</sup>
Just because I wrote it doesn't mean it's right :) I believe if this falls through it will return 500 Internal Server error, whereas it should return 406 Not Acceptable.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 54 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TvlsPqg50FS7e_ZJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L54)):</sup>
We might want to try [Babel](https://github.com/hammerlab/pileup.js/issues/21) instead of react-tools at some point—it claims to have significantly better support for ES6 than react-tools & supports JSX.

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 55 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L55)):</sup>
I think it's fine as you've written it now. Just to explain my thinking, though, there were two arguments:

- performance -- if we start using [PureRenderMixin](http://facebook.github.io/react/docs/pure-render-mixin.html) at some point, creating a new closure on every render will be a problem.
- [[define display](https://www.google.com/search?q=define+display)]
  > dis·play
  > verb
  > 1. make a prominent exhibition of (something) in a place where it can be easily seen. 

  So as a verb, having a method called display that actually hides something doesn't make much sense.

---

<sup>**[cycledash/validations.py, line 70 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L70)):</sup>
And it's been confusing me for months. :) I didn't actually realize that voluptuous had nothing to do with databases until I did the tasks change.

I generally don't find the Argument from Inertia / Appeal to Tradition very compelling -- there's nothing special about code we wrote a few months ago. It can always get better!

---

<sup>**[tests/python/test_api.py, line 9 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L9)):</sup>
I don't care too much in this case since the helper is so simple. But the general idea is that if you write a testing helper method, then you really need to unit test it (otherwise your tests are built on sand!). The API method is tested; this helper isn't. You could keep the helper function and just have it make an API call—there'd be no more or less code duplication.

---

<sup>**[tests/python/test_bams_api.py, line 41 \[r10\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk5Po-HdJA3nORh85eb)** ([raw file](https://github.com/hammerlab/cycledash/blob/645fac107862c5551e3b6d8a441aeab2dce4c7de/tests/python/test_bams_api.py#L41)):</sup>
I think the deleted fields were still in the revision that I saw. This is fine.

---

ihodes:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

Sweet, thanks for the review. Good to merge? 

I need to deal with some merge issues, hopefully not many, and finish writing the migration SQL up.

---

<sup>**[cycledash/bams.py, line 65 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4JhQs-U4C0ngzKqdQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/4ef0c6d719619c830a456b849af137910ae9fe09/cycledash/bams.py#L65)):</sup>
I agree—however, it shouldn't hold this PR up! 

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 56 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4TvlsPqg50FS7e_ZJ)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L56)):</sup>
Agreed https://github.com/hammerlab/cycledash/issues/536

---

<sup>**[cycledash/static/js/runs/components/RunsPage.js, line 57 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4Tyqq4p-cwghWSYX4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/static/js/runs/components/RunsPage.js#L57)):</sup>
Hah, yes I'm aware of the definition of `define` :) 

display(false) vs display(true) seems quite unambiguous, but as I already wrote (in the other place we're talking about this method), I've already renamed it since you find it confusing!

Regarding performance: if it becomes a problem ever (seems unlikely with the current state of affairs!) we can change it.

---

<sup>**[cycledash/validations.py, line 69 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4VyncRazfzWj57E8h)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/cycledash/validations.py#L69)):</sup>
I also don't find logical fallacies compelling, I like to think of myself as a rational person as well :)  My point was it seemed like it wasn't confusing since it'd been unquestioned for the lifetime of this file, but apparently it was confusing! For clarity, it *is* a schema; schemas don't imply database, they define data type (including shape). Nevertheless, you have convinced me that the suffix is unnecessary, and I've already removed it, so we should be good to go.

---

<sup>**[tests/python/test_api.py, line 0 \[r5\]](https://reviewable.io:443/reviews/hammerlab/cycledash/510#-Jk4XzhldnvfxjeADmiF)** ([raw file](https://github.com/hammerlab/cycledash/blob/6cc850a54e28ec755e252ce7e8148b594984fd0a/tests/python/test_api.py#L0)):</sup>
> the general idea is that if you write a testing helper method, then you really need to unit test it (otherwise your tests are built on sand!).

Ah, that wasn't at all clear to me from your previous comments, and I agree with that in principle. In practice, as you said, this is quite simple. 

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)







---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/510)



---

Yep, looks good!

---

danvk:
Please track down why the gene annotations column disappeared from the screenshots.

ihodes:
That had to do with a caching issue in pyensembl. Fixed, but now I can't generate screenshot anymore.

danvk:
You can update the screenshots by cherry picking b0acc26940fce383bda74c2e7bc5e82e24a27c45. You should really add some comments to the test DB, though—there's no reason we should stop testing those scenarios.

ihodes:
I was planning on it! Was trying to figure out why dpxdt wasn't working first.

PR #526, opened 2015-03-09
Fix issue with Load IGV Link

### LINKED ISSUE #523: The "IGV: Load at Locus" link only jumps to locus
hitting the "(reload)" link will actually load the data.

danvk:
Fixes #523 

Here's what I think was happening: initially, the "Load" link has a react ID of something like `.foo.1` and the "jump" link is `.foo.3`. When you click the link, React does a DOM update and swaps the links (so that they become "Jump" and "reload"). Because the order has changed, the "Jump" link is now `.foo.1` and the "reload" link is now `.foo.3`. React seems to only trigger the true DOM event at this point, after doing an update cycle. So the link that's clicked is `.foo.1`, which is now the "Jump" link. I found this sequencing quite surprising (and hard to debug!).

There are two fixes in this change, either of which would have worked on its own:

1. Adding explicit React keys to each link.
2. Disabling the on-screen swap when you click a link.

I'm doing both because it's a bit jarring to see the link you just clicked change out from underneath you.

Includes a regression test.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/526)


hammer:
wow that got real

ihodes:
As discussed offline, may be better to go back to the `setInitialState` route instead of the `shouldComponentUpdate` route. Nice tests!

ihodes:
(Will rereview quickly when you get that change in, if it still looks like the right thing to do tomorrow)

danvk:
@ihodes I switched back to caching the initial `props.hasOpenedIGV` in `state` and made the test more realistic. It passes now & the code works.

ihodes:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/526)







---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/526#-Jk3hqSf624_QHaUiAIB)** ([raw file](https://github.com/hammerlab/cycledash/blob/fecd173f4a75086fbf2a68d11973c04cefd84713/cycledash/static/js/examine/components/CommentBox.js#L174)):</sup>
This is to prevent the links from changing in the comment box currently open?

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/526)







---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/526#-Jk3hqSf624_QHaUiAIB)** ([raw file](https://github.com/hammerlab/cycledash/blob/fecd173f4a75086fbf2a68d11973c04cefd84713/cycledash/static/js/examine/components/CommentBox.js#L174)):</sup>
Correct. Want to suggest an alternative wording for the comment to make that clearer?

---

PR #529, opened 2015-03-10
More aggressive linting: console and unused vars

### LINKED ISSUE #528: Use linter to prevent console.log
It's easy for these to creep into code reviews, it would be helpful to automatically prevent it. It should be possible to make jshint yell about `console.log` statements by setting its [devel][1] option.

[1]: http://jshint.com/docs/options/#devel

danvk:
Fixes #528 

(first commit) The `console.log` statements I'm happy to see go. I converted the others into `throw`. LMK what you think.

(second commit) I've also set `"unused": "vars"`, which warns about unused variables, but not unused function parameters. This helped to eliminate a ton of dead code & unnecessary imports.

 cc @tavinathanson 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/529)

### DIFF
--- a/cycledash/static/js/examine/RecordStore.js
+++ b/cycledash/static/js/examine/RecordStore.js
@@ -379,7 +379,7 @@ function createRecordStore(run, igvHttpfsUrl, dispatcher, opt_testDataSource) {
     var search = window.location.search.substring(1),
         vars = search.split('&');
     var val = _.first(_.filter(vars, v => {
-      var [key] = v.split('=');
+      var key = v.split('=')[0];
       return decodeURIPlus(key) == name;
     }));
     if (val) {
--- a/cycledash/static/js/examine/RecordStore.js
+++ b/cycledash/static/js/examine/RecordStore.js
@@ -379,7 +379,7 @@ function createRecordStore(run, igvHttpfsUrl, dispatcher, opt_testDataSource) {
     var search = window.location.search.substring(1),
         vars = search.split('&');
     var val = _.first(_.filter(vars, v => {
-      var [key, val] = v.split('=');
+      var [key] = v.split('=');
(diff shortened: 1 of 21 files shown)


danvk:
Feel free to tell me to hold off on the second commit (unused vars) if it's going to be a merge nightmare for you.

danvk:
... and this is up-to-date again

cycledash/static/js/examine/RecordStore.js
```diff
@@ -379,7 +379,7 @@ function createRecordStore(run, igvHttpfsUrl, dispatcher, opt_testDataSource) {
     var search = window.location.search.substring(1),
         vars = search.split('&');
     var val = _.first(_.filter(vars, v => {
-      var [key, val] = v.split('=');
+      var [key] = v.split('=');
```
ihodes: This doesn't seem like a clear change (nor the one above with `[subquery]`; there should be some indication that we are indeed unpacking two values (like we'd use a `_` in Python), even if we're only using one of them.

ihodes:
LGTM % 1 comment

cycledash/static/js/examine/RecordStore.js
danvk: Would you be happy with

```javascript
  var [key /*, value */] = v.split('=');
```

???

ihodes: I'm not sure what the best choice is here; in other languages leaving out the second var to be unpacked would result in an error. Maybe the best choice is to turn off this lint rule? Could also do 

```javascript
var key = v.split('=')[0];
```

danvk: I like it, done.

PR #538, opened 2015-03-13
Add combobox selection for VCF to validate against

ihodes:
Also start the slow process of renaming run -> vcf.

pdiff tests.yaml is updated, and works in Chrome (and is purportedly supported in PhantomJS?) but isn't generating my expected screenshot. Mind giving it a try?

fixes #520 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/538)

PR #539, opened 2015-03-13
Display FILTER & QUAL fields in the table

### LINKED ISSUE #473: Display FILTER field in a VCF
For an example VCF
```
chr1    12907620        .       A       G       .       PASS    ....
chr1    12920828        .       C       A       .       QSS_ref ...
```

The 7th field is the `FILTER` field which is not visible in Cycledash currently

ihodes:
Fixes #473 

Pending #538 

Unfortunately, I don't have a run with filters in my table, so we see an empty column in the VCF. I'll file an issue for this.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/539)


danvk:
LGTM, feel free to merge on green.

PR #548, opened 2015-03-17
 seltest pdiff tests

ihodes:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/548)


danvk:
API generally seems reasonable -- I like the `@waitFor`.

Like I mentioned yesterday, we should try to keep the test outputs as similar as possible: same file name, same dimensions, same content. This will let us focus in on finer changes & spot small regressions.

danvk:
Can we keep the image file names the same? It's hard to track the changes if they're also moves.

ihodes:
They will remain the same from here on out--in fact, nearly all of them are exactly the same. I've just moved them all into the `images` subdirectory in order to keep things cleaner. I believe that you'll find that there are no regressions.

danvk:
Is the switch from 1200px→2000px viewport intentional? (As in, you're explicitly saying testing on a wider browser is a good idea.)

I do feel sad about losing version history on the screenshots files.

I'll try to generate screenshots locally.

tests/pdifftests/run.sh
```diff
@@ -0,0 +1,8 @@
+source venv/bin/activate
```
danvk: Add a shebang line

danvk: Remove this -- it will only work on your machine.

danvk:
A few more issues from in-person discussion:

- Put mouse cursor in a set position at the start of every test.
- Run tests in a deterministic order to reduce cross-pollution.
- Don't write PNG headers—they include timestamps & are a source of differences between runs.

ihodes:
Alright, added those features/fixes to seltest! Give it another try? I need to upgrade my OS to match the UI in 10.10.2.

danvk:
re: version history on these screenshots, how about I send out a PR that's a simple rename & we merge it in before this? That way we should be able to keep the revision history of the screenshots across this change.

ihodes:
@danvk I'm not sure it'll be that useful for us in practice, but I certainly don't object to it!

ihodes:
This should be ready to merge now. I switched the default test runner to using phantomjs in the interim, so that our test output will look the same across computers and minimize the change in workflow right now, but the runner will run with any browser. 

In a future PR, I will integrate this with SauceLabs to run tests on multiple browsers as part of the TravisCI suite. 

A nice side-effect of this change is that the tests, though there are a couple more of them, run in less than half the time they did before. 

@danvk back to you!

danvk:
If we're using Phantom for the screenshots, we shouldn't see many differences from dpxdt then, right?

Does Selenium run multiple tests in parallel with Phantom? I'm a bit surprised you're seeing such a performance difference from dpxdt.

ihodes:
> If we're using Phantom for the screenshots, we shouldn't see many differences from dpxdt then, right?

Correct.

> Does Selenium run multiple tests in parallel with Phantom? I'm a bit surprised you're seeing such a performance difference from dpxdt.

It does not; I have [an issue filed](https://github.com/ihodes/seltest/issues/40) to run them in parallel, though, which should speed things up significantly, especially as we add more tests!

danvk:
Can you try generating 1200px wide screenshots? I'd like to see comparable before/after dpxdt/seltest diffs.

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/548)



---

Nice—these diffs look pretty reasonable! Comments on them here since you can't comment directly on image diffs:

- `examine_bad-query.png` Your version looks fine, but why was the query bad before?
- `examine_sorted.png` Any reason to change the sort order?
- `examine_tooltip.png` Arbitrary change to the highlighted column?
- `runs_info.png` Can you either black out the "X days ago" text or replace all the numbers with some constant in your Python code? Otherwise they're going to change every time we update the screenshots.

---

<sup>**[DEVELOP.md, line 169 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7fj7U-to1L7OgAxz4)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/DEVELOP.md#L169)):</sup>
This isn't correct anymore

---

<sup>**[tests/pdifftests/README.md, line 4 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7hFvuqWQ8LLwQchvu)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/README.md#L4)):</sup>
dpxdt → image diff (or whatever we want to call them)

---

<sup>**[tests/pdifftests/README.md, line 11 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7hMVanJ5GyWK9ar8K)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/README.md#L11)):</sup>
also here

---

<sup>**[tests/pdifftests/run.sh, line 1 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7hQOr_1bzr1oi2Z0L)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/run.sh#L1)):</sup>
Should probably just use `/bin/bash`—would this script work if the shell were real `sh` or `tcsh`?

---

<sup>**[tests/pdifftests/run.sh, line 8 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7g-r1aFvZsRuC0KiD)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/run.sh#L8)):</sup>
In general you're better off doing this with an [exit trap](http://redsymbol.net/articles/bash-exit-traps/). It'll let you `set -o errexit` and still be confident that you'll kill any servers you start.

`set -o errexit` is probably a good idea regardless, since you wouldn't want to run `sel` if the `python run.py` step fails.

---

<sup>**[tests/pdifftests/test.py, line 66 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7gcI9zUVXp4M5pU-F)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/test.py#L66)):</sup>
for bonus points, run this through a URL encoding library to get totally readable CQL!

---

<sup>**[tests/pdifftests/worker.sh, line 4 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7htQO2ou88s0uIdmR)** ([raw file](https://github.com/hammerlab/cycledash/blob/04f3df1adce42e784cce112be3080b74ba113f57/tests/pdifftests/worker.sh#L4)):</sup>
is this not useful any more?

---

ihodes:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/548)



---

* I don't recall!
* I don't think so.
* Yes.
* That makes sense to me; I'll file an issue for that!

---

<sup>**[tests/pdifftests/run.sh, line 1 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7hQOr_1bzr1oi2Z0L)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/run.sh#L1)):</sup>
It should! I'll switch it.

---

<sup>**[tests/pdifftests/run.sh, line 8 \[r7\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7g-r1aFvZsRuC0KiD)** ([raw file](https://github.com/hammerlab/cycledash/blob/6e88a902d30ef4e164e801190cdcbc1c592f1a0c/tests/pdifftests/run.sh#L8)):</sup>
I purposefully omitted `set -e` as I like to have my own test server running on 5001 sometimes, so I'd rather have it continue.

I'll add the exit trap!

---

danvk:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/548)



---

May as well change them back to keep continuity in the tests, then. (`sorted` and `tooltip`)

---

<sup>**[tests/pdifftests/README.md, line 1 \[r8\]](https://reviewable.io:443/reviews/hammerlab/cycledash/548#-Jl7ksZ7ggOH581IPg12)** ([raw file](https://github.com/hammerlab/cycledash/blob/cdd9b0f46dcfc750e48d744b84d32fb4cfa30de5/tests/pdifftests/README.md#L1)):</sup>
image diff → perceptual diff

(not your writing, but we may as well use consistent terminology)

---

ihodes:
> May as well change them back to keep continuity in the tests, then. (sorted and tooltip)

Looks like `sorted` was previously sorting by the contig column, which is sorted by by default anyway. Makes sense to change it to test for something different.

I will move the tooltip back to the other column.

Issue #569, opened 2015-04-01
"Last 5 Comments" has a text wrap issue in the pdifftest screenshots


tavinathanson:
@danvk noticed that there's an issue with text wrap in the "Last 5 Comments" table showing up in runs_info.png.

![image](https://cloud.githubusercontent.com/assets/322932/6946515/dda467de-d86e-11e4-9ea9-9efdccc36c2e.png)

ihodes:
These wrap more appropriately now (cf. current https://raw.githubusercontent.com/hammerlab/cycledash/master/tests/pdifftests/images/runs_page.png); we also have https://github.com/hammerlab/cycledash/issues/650 to track improving this.

PR #572, opened 2015-04-01
Fix feed times bug and use created_date

tavinathanson:
Subtracting timezoneOffset (number) from a timestamp didn't make sense (also, should be adding to GMT to get local...), so I changed that to use a Moment function (`.add`). I also swapped last_modified for created.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/572)


tavinathanson:
Looks like https://github.com/hammerlab/cycledash/issues/569 isn't present here, so that appears to be a random issue that surfaces only sometimes :( cc @danvk

ihodes:
It'll likely only occur when the timestamp is "a few seconds ago"—the rest are too short to cause the overflow, I'd hazard.

LGTM!

tavinathanson:
@ihodes Ah, good point!

PR #573, opened 2015-04-01
Fix comment sorting bug

tavinathanson:
Missing return!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/573)


ihodes:
LGTM—I think this is a bug in our transpiler; I thought the last statement in an ES6 fat-arrow lambda should return?

PR #576, opened 2015-04-02
Fix timezones

tavinathanson:
From http://www.postgresql.org/docs/9.1/static/datatype-datetime.html: "For `timestamp with time zone`, the internally stored value is always in UTC (Universal Coordinated Time, traditionally known as Greenwich Mean Time, GMT)."

Versus, for `timestamp without time zone`: "a literal that has been determined to be `timestamp without time zone`, PostgreSQL will silently ignore any time zone indication. That is, the resulting value is derived from the date/time fields in the input value, and is not adjusted for time zone."

Other information seems to agree with just using `timestamp with time zone` (`timestampz`): http://justatheory.com/computers/databases/postgresql/use-timestamptz.html

So, we now use `timestamp with time zone` + set UTC in the DB's conf. (Though, that UTC setting in the conf file doesn't appear to be necessary, per: "it is assumed to be in the time zone indicated by the system's `timezone` parameter, and is converted to UTC using the offset for the `timezone` zone".) Then, moment takes that UTC timestamp and displays the localized version of it. No offsets necessary.

@danvk regarding microsecond precision (when, according to the documentation, there don't appear to be enough digits to support that): "timestamp values are stored as seconds before or after midnight 2000-01-01. When timestamp values are implemented using floating-point numbers, microsecond precision is achieved for dates within a few years of 2000-01-01, but the precision degrades for dates further away. Note that using floating-point datetimes allows a larger range of timestamp values to be represented than shown above: from 4713 BC up to 5874897 AD."

Migration code:

```
alter table user_comments alter column created type timestamp with time zone;
alter table user_comments alter column last_modified type timestamp with time zone;
```

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/576)


danvk:
Looks fine, thanks for trudging through this!

My question about the microseconds was why the the interval isn't symmetric around the epoch (in this case midnight 2000-01-01). 4713BC is way closer to 2000 than 5874897 AD is!

---



---

<sup>**[cycledash/comments.py, line 254 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-JlvzgnbJycdFtKaEfd4)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/comments.py#L254)):</sup>
Can you use [this](http://www.postgresql.org/message-id/1046309418.88663.8.camel@jester) to extract seconds since epoch straight out of postgres?

---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-AWUxNwhq1oQDD16)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/static/js/examine/components/CommentBox.js#L107)):</sup>
What is `created_date`? Is it seconds since epoch? Calculating elapsed time from a timestamp (i.e. seconds since epoch) shouldn't require any knowledge of time zones.

---

<sup>**[tests/data/comment-db.json, line 7 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-XyTPfxbW3BcBkOQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/tests/data/comment-db.json#L7)):</sup>
FYI `Z` == `+00:00`

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

tavinathanson:
Ah, interesting re the asymmetry.

---



---

<sup>**[cycledash/comments.py, line 254 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-JlvzgnbJycdFtKaEfd4)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/comments.py#L254)):</sup>
I was playing around with that very query for something else! In any event, that just gives seconds, though there's probably another query I can use for milliseconds. Stay tuned.

---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-AWUxNwhq1oQDD16)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/static/js/examine/components/CommentBox.js#L107)):</sup>
The API currently returns ISO 8601 format: "2015-03-10T19:09:49.170964+00:00", which is what gets passed into moment. I wanted to the comment to clarify that this timestamp was UTC, but conversion to local is handled automatically.

You might also be wondering why `created_date` is ISO 8601 while `last_modified_timestamp is microseconds` since epoch. The short answer is that they're doing different things (one is for display and one is for handling clobbering, and their different formats are sort of reflected in `date` vs. `timestamp`).

That said, that's confusing, and I'd be up for switching `created_date` to microseconds since epoch (or seconds since epoch with a floating point, whichever is easiest). I'll look at that now.

---

<sup>**[tests/data/comment-db.json, line 7 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-XyTPfxbW3BcBkOQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/tests/data/comment-db.json#L7)):</sup>
Gotcha, though not sure how to use this information?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

danvk:
---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-AWUxNwhq1oQDD16)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/static/js/examine/components/CommentBox.js#L107)):</sup>
Yeah, my preference would be to have all dates coming from the backend be represented in one format. I'd prefer that format to be seconds since epoch (floating point) if that's feasible.

---

<sup>**[tests/data/comment-db.json, line 7 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jlw-XyTPfxbW3BcBkOQ)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/tests/data/comment-db.json#L7)):</sup>
Just an FYI, no need to do anything :)

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

tavinathanson:
---

<sup>**[cycledash/comments.py, line 254 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-JlvzgnbJycdFtKaEfd4)** ([raw file](https://github.com/hammerlab/cycledash/blob/a4acb00a10c645e0a5e477af8164b020980f615b/cycledash/comments.py#L254)):</sup>
Ah, I was wrong. That does extract microseconds after a floating point.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

tavinathanson:
I went ahead and made `created`/`last_modified` both seconds (.microseconds) since epoch. As you suggested, I was able to use the SQL itself to get the epoch (though it involved a bit of SQLalchemy column set fiddling). Overall, I think `comments.py` is simpler now (without the need for any of the date_as_int stuff).

On the JS side, `created` and `last_modified` are now both suffixed with `_timestamp`.

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

tavinathanson:
I also decided to sort the feed by `created_timestamp` to be in line with the date shown on each comment, though I didn't give that too much thought and could go either way. (Screenshots updated accordingly.)

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

danvk:
Seems good!

---



---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 93 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jm049JiYXDAAguNyZ8i)** ([raw file](https://github.com/hammerlab/cycledash/blob/30a40a4fec21e19e24b4d4e9b52501f1a7fe8bdd/cycledash/static/js/examine/components/CommentBox.js#L93)):</sup>
I'm having trouble reconciling this with the simple one-liner below. :) Maybe I'm missing what the hard bit is?

---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 119 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jm04mi2xbhyRc-v4FL_)** ([raw file](https://github.com/hammerlab/cycledash/blob/cd0615d663b291d27c1e51946184f06eda6adf68/cycledash/static/js/examine/components/CommentBox.js#L119)):</sup>
Big nit! General rule of thumb for wrapping to 80 character lines is to break on the highest level of the AST that you can. (My only reference for that is an internal Google rule, but it holds up well in practice.) So I'd probably put the break before `String` on this line (if it fits), or maybe after `String(` if it doesn't.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

tavinathanson:
Good to merge?

---



---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 93 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jm049JiYXDAAguNyZ8i)** ([raw file](https://github.com/hammerlab/cycledash/blob/30a40a4fec21e19e24b4d4e9b52501f1a7fe8bdd/cycledash/static/js/examine/components/CommentBox.js#L93)):</sup>
I agree that it's over-commented! The function started out a bit more complicated. I shortened the comment :)

---

<sup>**[cycledash/static/js/examine/components/CommentBox.js, line 119 \[r3\]](https://reviewable.io:443/reviews/hammerlab/cycledash/576#-Jm04mi2xbhyRc-v4FL_)** ([raw file](https://github.com/hammerlab/cycledash/blob/cd0615d663b291d27c1e51946184f06eda6adf68/cycledash/static/js/examine/components/CommentBox.js#L119)):</sup>
Hmm. I appreciate the comment because I haven't resolved my own style preferences for wrapping. I also have a tendency to wrap things as I would wrap them in Python, for better or for worse. That said, the break before `String` or after `String(` results in going past 80 chars. So I'll break on `.getGranularUnixSeconds(`?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/576)

PR #581, opened 2015-05-06
Updates to DEVELOP.md

jaclynperrone:
Updated DEVELOP.md based on my experience running CycleDash for the first time. Changed the order of a couple of things and made it read more like steps. Removed a couple of duplicate lines, and added a few notes that I think may be helpful for first time users.

Feel free to make any changes! Would love your feedback and thoughts on how we could consolidate it even more.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/581)


DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
+pip install virtualenv
+```
+
+
+### 2. Set up CycleDash
 
 ```bash
 virtualenv venv                    # Initialize a new virtual environment.
-source venv/bin/activate           # Activate your virtual environment.
+source venv/bin/activate           # Activate your virtual environment. This is the only command you will need to run every time you start the app.
```
danvk: Instead of noting this in a comment, how about adding a second paragraph like:

After the initial setup, the only command you'll need to run is:

```bash
source venv/bin/activate
```

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
 pip install -r requirements.txt    # Install python requirements into virtualenv
-npm install                        # Install JS requirements
+npm install                        # Install JS requirements and packages in package.json
```
danvk: The "JS requirements" are "packages", so how about just

`# Install JS requirements from package.json`

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
-### Database
+### 3. Edit the ENV.sh file (generated by `make initenv`)
 
-We use PostgreSQL as our datastore, with the schema described in `schema.sql`.
-
-You can load the schema with `psql cycledash < schema.sql`, and you'll be good
-to go.
-
-On OS X, setting up and running psql might look like this:
+Environment variables which must be exported--edit them in the ENV.sh file make
```
danvk: sentence fragment, reword

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
-```
+We use PostgreSQL as our datastore, with the schema described in `schema.sql`.
 
-Start a worker to process the queue. Make sure RabbitMQ is up and running!
+On OS X, setting up and running psql might look like this:
 
+```bash
+postgres -D /usr/local/var/postgres
+createdb cycledash
+psql cycledash
```
danvk: @ihodes this command isn't necessary, is it?

danvk:
This looks good, just a few comments. Thanks for making these changes @jaclynperrone!

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
 4. [RabbitMQ](http://www.rabbitmq.com/) for workers
 
 
-### Setting up CycleDash
+### 1. Install Dependencies
+
+```bash
```
ihodes: I'd note that is for a Mac install, and that other OSs would have similar procedures (using e.g. apt-get or yum).

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
+Environment variables which must be exported--edit them in the ENV.sh file make
+made for you.
 
 ```
-brew install postgres
-postgres -D /usr/local/var/postgres
-createdb cycledash
-psql cycledash
-psql cycledash < schema.sql
+export PORT=5000
```
ihodes: May be worth noting or copying from ENVTEMPLATE.sh, which has these default values I think?

DEVELOP.md
ihodes: Correct, this spins up the postgres shell.

DEVELOP.md
```diff
@@ -9,65 +9,66 @@ Dependencies:
...
+### 5. Set up the JavaScript
 
 You can iterate more quickly on JavaScript with the following:
 
 ```
-npm install             # Installs all packages in package.json.
 npm install gulp -g     # Make sure you have gulp installed.
-gulp                    # Compile the JS and start the automatic compiler
-                        # and live-reloader.
+gulp                    # Compile the JS and start the automatic compiler and live-reloader.
```
ihodes: I'd keep this on two lines; under the 80 char archaic line length limit we try to stick to.

DEVELOP.md
```diff
@@ -90,37 +91,44 @@ Run `gulp peg` to update the PEG.js generated grammar after modifying CQL.
...
 ```bash
-brew install rabbitmq
+gulp prod
+./run.sh
+```
+
+
+### 7. Run RabbitMQ (and keep it running)
+
+To test the workers locally, you'll need to install and run rabbitmq during development. For example, on Mac OS X, you can do this via:
```
ihodes: Similarly, try to stick to the 80 char line limit.

(Or else we could switch the entire file to use no line length limit, instead?)

ihodes:
Thanks for fixing this up! Looks good, few comments.

DEVELOP.md
jaclynperrone: Noted & added!

DEVELOP.md
jaclynperrone: Excellent. Added.

DEVELOP.md
jaclynperrone: @danvk Would something like this make sense: "ENV.sh is generated by 'make initenv' and contains environment variables that need to be exported."

Not sure if the "contains environment variables that need to be exported" bit makes sense?

danvk: :+1:

DEVELOP.md
jaclynperrone: Totally! Just talked with @danvk and we agreed to remove most of this section. I updated the DATABASE_URI value in ENVTEMPLATE.sh to "postgres:///cycledash"

DEVELOP.md
jaclynperrone: noted!

DEVELOP.md
jaclynperrone: Noted!

DEVELOP.md
jaclynperrone: Sure thing. I added a line about this.

DEVELOP.md
jaclynperrone: It's a little tough to see which line this comment is pointing to, but I am assuming it's
```
psql cycledash
```

danvk:
Just two nits, otherwise looks good!

danvk:
Looks good—go ahead and hit the green Merge button!

Issue #574, opened 2015-04-02
IGV integration broken


danvk:
BAM tracks wouldn't load during the demo w/ the plot.ly folks today. Some kind of regression?

arahuja:
I couldn't find a working example today - was this a real issue?

danvk:
Yeah, it is. I'll debug now.

danvk:
http://cycledash.demeter.hpc.mssm.edu/runs/270 illustrates some of the problems.

One issue is that the VCF file for that run doesn't end in `.vcf` (it's `/foo.vcf/part-r-00000`). IGV switches track type on the file extension, so it has no idea what it's looking at.

I'm not sure why the BAM files aren't being included in the URL yet.

danvk:
The issue with BAMs is that `run.normal_bam.uri` and `run.normal_bam.uri` got dropped at some point and none of the tests were strong enough to catch it. @ihodes does that sound familiar? Maybe part of the reorganization of the /runs page?

ihodes:
Probably so; I don't recall the particulars.

danvk:
The /runs page generates a "run" object through `_get_projects_tree`.
The /examine page generates it through `runs.get_vcf`.

One solution is to have `runs.get_vcf` call `_get_projects_tree` and find the relevant VCF. This will pull in more data than is necessary, but it will result in a more uniform run/vcf object.

danvk:
#591 fixes the missing BAM issue.

The only workaround I can see for the VCF issue is to serve the VCF through cycledash itself with a different suffix. @arahuja would requiring that VCFs have the right extension be an undue burden?

danvk:
A separate issue is that IGV requires an index for sufficiently large VCF files. This happens on run 189, for instance, where it issues a request for the non-existent  `http://hammerlab-dev3.hpc.mssm.edu:9876/datasets/martignetti_ovarian/189/Ion_Exome/TSVC_variants_IonXpress_019.vcf.idx`. The VCF is 30+ MB. I wasn't aware that there was a standard index format for VCFs!

danvk:
See https://github.com/igvteam/igv/issues/146

danvk:
The IGV people are going to fix the VCF extension issue upstream. Once that happens I'll use the fix & we can close this out.

arahuja:
To be honest, having IGV show the variant is not that critical - since we are only selecting positions where there is a variant, the additional box that signifies it doesn't add much. 

[...]

PR #591, opened 2015-05-07
Unify /runs and /examine vcf object

danvk:
See #574 

Previously the `vcf` object on the /examine page only had tumor_bam_id and not tumor_bam_uri. There were a few other subtle distinctions between the examine page's `vcf` object and the equivalent on the runs page.

This results in more work being done on the initial /examine request, but it's an exact copy of the /runs request, so hopefully it's not a big deal.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/591)

### DIFF
--- a/cycledash/bams.py
+++ b/cycledash/bams.py
@@ -19,14 +19,14 @@ def attach_bams_to_vcfs(vcfs):
     with tables(db, 'bams') as (con, bams):
         q = select(bams.c)
         bams = [dict(b) for b in con.execute(q).fetchall()]
-        for vcf in vcfs:
-            normal_bam_id = vcf.get('normal_bam_id')
-            tumor_bam_id = vcf.get('tumor_bam_id')
+    for vcf in vcfs:
+        normal_bam_id = vcf.get('normal_bam_id')
+        tumor_bam_id = vcf.get('tumor_bam_id')
 
-            vcf['tumor_bam'] = (
-                dict(find(bams, lambda x: x.get('id') == tumor_bam_id) or {}))
-            vcf['normal_bam'] = (
-                dict(find(bams, lambda x: x.get('id') == normal_bam_id) or {}))
+        vcf['tumor_bam'] = (
+            dict(find(bams, lambda x: x.get('id') == tumor_bam_id) or {}))
+        vcf['normal_bam'] = (
+            dict(find(bams, lambda x: x.get('id') == normal_bam_id) or {}))
 
 
 def get_bam(bam_id):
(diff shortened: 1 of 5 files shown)


cycledash/runs.py
```diff
@@ -21,9 +21,15 @@ class CollisionError(Exception):
 def get_vcf(vcf_id):
     """Return a vcf with a given ID, and the spec and list of contigs for that
     run for use by the /examine page."""
-    with tables(db, 'vcfs') as (con, vcfs):
-        q = select(vcfs.c).where(vcfs.c.id == vcf_id)
-        vcf = dict(con.execute(q).fetchone())
+    projects = cycledash.projects.get_projects_tree_dict()
+    vcf = None
+    for project in projects:
```
ihodes: I think it'd be better to just request the project & BAM objects for this particular VCF rather than request it for all VCFs' BAMs and Projects and throw away all the extra information.

ihodes:
Thanks for handling this; back to you!

cycledash/runs.py
danvk: I'd rather not duplicate code and have the objects diverge again. As this issue indicated, it's a source of bugs.

Is the `get_projects_tree_dict()` call particularly expensive?

ihodes: The objects need to be different; I should've noticed this before, but does this fix the problem? BioDalliance is passed `vcf.normal_bam_url`, which still isn't present when you do this. 

A new function is called for, which just adds the BAM URIs to the VCF object, for example.

ihodes: cf https://github.com/hammerlab/cycledash/blob/master/cycledash/static/js/examine/components/ExaminePage.js#L109

danvk: Good catch. I've updated the references. BAMs are displaying in dalliance once again!

danvk:
I factored out `cycledash.bams.attach_bams_to_vcfs`. Let me know if this is what you had in mind.

cycledash/bams.py
```diff
@@ -14,6 +14,21 @@
 import workers.indexer
 
 
+def attach_bams_to_vcfs(vcfs):
+    """Attaches tumor_bam and normal_bam to all the VCFs."""
+    with tables(db, 'bams') as (con, bams):
+        q = select(bams.c)
+        bams = [dict(b) for b in con.execute(q).fetchall()]
+        for vcf in vcfs:
```
ihodes: NBD, but could unindent here so that the DB connection can be closed.

ihodes:
Yeah, that was my thought. LGTM! 1 minor comment.

cycledash/bams.py
danvk: Done

danvk:
I'll merge on green.

Issue #537, opened 2015-03-12
Fix all Postgres connection leakes


ihodes:
Should probably be using `tables` or another context manager when using the database.

I'm guessing the leaks are in the workers?

ihodes:
Difficult to tell, but some testing seems to indicate that there are 4 connections leaking every time `gene_annotator.annotate` is run. I'm not very confident in this finding, though.

PR #594, opened 2015-05-08
Add a "restart failed tasks" button to the tasks page

ihodes:
Now failed tasks can be restarted from the web UI

Also simplified code in gene_annotator in pursuit of a postgres
connection link (#537). May have gotten rid of one or two, but some
appear to remain.

Got rid of a DB leak when tasks' status were refreshed.

Used tables(..) context manager in workers to simplify code and safely
close connections.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/594)

### DIFF
--- a/common/helpers.py
+++ b/common/helpers.py
@@ -5,16 +5,16 @@ import sqlalchemy
 
 
 @contextmanager
-def tables(database, *table_names):
+def tables(engine, *table_names):
     """A context manager yielding a tuple of the database connection and
     whichever tables were requested by name from the db.
 
     Use:
         with tables(db, 'vcfs', 'genotypes') as (con, vcfs, genotypes):
             ...
-      Where vcfs and genotypes are tables in the provided db.
+        Where vcfs and genotypes are tables in the provided db.
     """
-    with database.engine.connect() as connection:
+    with engine.connect() as connection:
         metadata = sqlalchemy.MetaData(bind=connection)
         metadata.reflect()
         yield tuple([connection] + [metadata.tables[t] for t in table_names])
--- a/common/relational_vcf.py
+++ b/common/relational_vcf.py
@@ -133,10 +133,10 @@ def vcf_to_csv(vcfdata, columns, filename, **kwargs):
 
 def insert_csv(filename, tablename, engine):
     """Insert the relations in the CSV filename into table, using engine."""
-    with engine.raw_connection() as raw:
-        cur = raw.cursor()
-        cur.execute("COPY {} FROM '{}' WITH (FORMAT csv);".format(tablename, filename))
-        raw.commit()
+    raw = engine.raw_connection()
+    cur = raw.cursor()
+    cur.execute("COPY {} FROM '{}' WITH (FORMAT csv);".format(tablename, filename))
+    raw.commit()
 
 
 def insert_genotypes_with_copy(vcfreader, engine, **kwargs):
--- a/common/relational_vcf.py
+++ b/common/relational_vcf.py
@@ -133,12 +133,10 @@ def vcf_to_csv(vcfdata, columns, filename, **kwargs):
 
 def insert_csv(filename, tablename, engine):
     """Insert the relations in the CSV filename into table, using engine."""
-    raw = engine.raw_connection()
-    cur = raw.cursor()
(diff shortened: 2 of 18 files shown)


danvk:
This mostly looks fine. Two high level comments, one inline, one below...

Is there a good way to automatically check for leaked connections? This is a good thing to wrap around tests. For example, Mocha checks for leaked global variables by diffing properties on `window` before and after you run your tests. Perhaps we could track opening/closing DB connections and make sure opens==closes after tests run. There may be tooling to help with this. I'm sure we've all introduced leaks without realizing it. If we don't automate the checks, we'll probably do it again in the future.

---

Reviewed files:
- common/helpers.py @ r1
- common/relational_vcf.py @ r1
- cycledash/bams.py @ r1
- cycledash/comments.py @ r1
- cycledash/genotypes.py @ r1
- cycledash/helpers.py @ r1
- cycledash/projects.py @ r1
- cycledash/runs.py @ r1
- cycledash/tasks.py @ r1
- cycledash/templates/tasks.html @ r1
- cycledash/views.py @ r1
- requirements.txt @ r1
- tests/python/test_bams_api.py @ r1
- tests/python/test_projects_api.py @ r1
- workers/gene_annotator.py @ r1
- workers/genotype_extractor.py @ r1
- workers/runner.py @ r1
- workers/shared.py @ r1

---

<sup>**[common/helpers.py, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/594#-Jp2YF4n9lnMYMXw2HXs)** ([raw file](https://github.com/hammerlab/cycledash/blob/68051bcccfaf047d0a8777a2d921eab4a5f914d2/common/helpers.py#L8)):</sup>
I'm not sure I understand the need for this change. It look like all the server code has a reference to `db` and passes in `db.engine`, whereas all the worker code just has a reference to `engine`. Is there a reason that server code & worker code get their database connections in different ways? Can you unify them?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/594)

ihodes:
> Is there a good way to automatically check for leaked connections? 

I looked into this a bit, but I couldn't find anything after a bit of searching. I haven't given up all hope, though.

One issue is we don't necessarily want `.close()` called on all connection as soon as they're used; we use threadpools so that we don't have to do expensive Postgres connections (they're over TCP, and require other database-specific negotiation steps as well) every time we want to call out to the DB.

I think using context managers where possible would *mostly* fix this. And perhaps better use of threadpools in the worker code would help as well. 

The postgres connection issue is definitely still open, and bears more investigation. This will just help things a bit, I hope.

---





<sup>**[common/helpers.py, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/cycledash/594#-Jp2YF4n9lnMYMXw2HXs)** ([raw file](https://github.com/hammerlab/cycledash/blob/68051bcccfaf047d0a8777a2d921eab4a5f914d2/common/helpers.py#L8)):</sup>
> Is there a reason that server code & worker code get their database connections in different ways? 

Unfortunately, since we're using Flask-SQLAlchemy, we construct the DB via the Flask App (e.g. `db = SQLAlchemy(app)`) which reads the config and does a bunch of other tricky Flask-specific stuff + I think sets up a threadpool. In the workers, we don't spin up (and don't want to; and it'd require dealing with some weird circular imports) a Flask app, so we just construct the engine with vanilla SQLAlchemy. 

I think the ideal solution is probably to get rid of Flask-SQLAlchemy and use only vanilla SQLAlchemy. I think we'd need to handle setting up a shared threadpool and sharing it safely across Flask instances, but I'm not really sure all that's involved. I don't think the situation we have now (using Flask-SQLAlchemy for the Flask app, and just SQLAlchemy for the non-Flask app stuff) is bad, though.

As a side note, I've come to avoid Flask plugins as a rule. They make things *slightly* easier to write at first, but they can cause issues later. Maybe there's a simpler way around this, but I think this solution is pretty straightforward. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/594)

danvk:
You've clearly thought about both these issues :)

Feel free to merge!

PR #596, opened 2015-05-12
Sass spike

jaclynperrone:
Installed sass and converted the css stylesheets into scss format. No changes have been made to the styles themselves.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/596)


ihodes:
Looks good! Couple comments + we should rope @danvk in and see what he thinks of sass as well. 

Main one is we need to delete the existing css files and add them to .gitignore so they don't show up in the repo anymore.

From a VC perspective; it'd be nice to `git mv` (and rename) the css files into the `scss` directory so that git is aware this change is a move.

jaclynperrone:
@ihodes  +1 on the `git mv` approach. I used it to rename and move the css files...but maybe I need to do something differently?

ihodes:
It may be that git is confused because the .css files are still in the repository (though now we're generating them with sass). I'd try adding them to `.gitignore` and see what git thinks happened? You may need to amend a commit—I'm not super sure about how git handles moves/deletes/rename in this case.

danvk:
I'm generally pro-sass. I agree with what Isaac says about structuring this as a pure move followed by a `.gitignore` change. Note that this isn't a pure move right now: there are lots of odd formatting changes.

jaclynperrone:
@danvk What's "pure move"?

danvk:
"pure move" = file content doesn't change. There are changes to these files, e.g. moving the `}` up to the previous line. git tries to detect moves based on file similarity, so rename + change can throw it off.

PR #597, opened 2015-05-13
Start using Sass

jaclynperrone:
Installed Sass and converted the css stylesheets into scss format. No changes have been made to the styles themselves.

Supersedes #596

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/597)


danvk:
FYI @jaclynperrone I figured out why your build is broken :)

danvk:
Travis runs `gulp prod` before starting. Previously this generated some files and then terminated. Now it does that _and_ starts the watch loop, i.e. it never terminates. So Travis is waiting for a process to terminate that never does, hence the timeouts.

`gulp sass` should terminate as it did before (and also generate the css files). Running plain-old `gulp` should do the watching.

jaclynperrone:
@danvk Ahhh, gotcha. Not sure which file I need to change. Does it have to do with the watch task in gulpfile.js?
screen shot 2015-05-13 at 5 07 14 pm

gulpfile.js
```diff
@@ -53,17 +54,19 @@ gulp.task('js', function() {
...
-
-// Notifies livereload server to reload the browser when CSS changes.
-gulp.task('css', function() {
-  livereload.changed();
+ 
+gulp.task('sass', function () {
+    gulp.src('./cycledash/static/scss/*.scss')
+        .pipe(sass().on('error', sass.logError))
+        .pipe(gulp.dest('./cycledash/static/css'));
+        livereload.changed();
```
danvk: Try removing this line.

jaclynperrone:
@danvk Would you mind giving me a line number?

PR #628, opened 2015-05-21
Extend Bootstrap grid and components to project.

jaclynperrone:
This commit features css refactoring in favor of @extending Bootstrap grid and components. Some things will look different (ie. buttons, selects, VCF table), but for the most part the original styling is intact. This makes it easier to merge it into master, for there aren't any major interface changes.

There are some markup changes in the .js files, but nothing more than adding/removing/renaming classes & elements.

The final thing to do is update all the screenshots, which I have not done yet.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/628)


ihodes:
Awesome! Can you run the pdiff tests on this?

cycledash/static/js/runs/components/forms.js
```diff
@@ -93,30 +100,37 @@ var NewBAMForm = React.createClass({
...
-        <h3>New BAM</h3>
-
-        <TextInput label='Name:' name='name' required={true}
-                   placeholder='...' />
-        <TextInput label='Tissues:' name='tissues'
-                   placeholder='Left Ovary' />
+      <form method='POST' action='/api/bams' className='add-form' ref='bamForm'>
+        <div className='row'>
+          <h3>New BAM</h3>
+          <div className='add-form-input-half'>
```
ihodes: Can we add the .add-form-input-half style to the underlying TextInput instead of wrapping each TextInput with this div?

cycledash/static/js/examine/components/ExaminePage.js
```diff
@@ -78,44 +78,57 @@ var ExaminePage = React.createClass({
...
-        <PileupViewer vcfPath={run.uri}
-                      normalBamPath={run.normal_bam && run.normal_bam.uri}
-                      tumorBamPath={run.tumor_bam && run.tumor_bam.uri}
-                      igvHttpfsUrl={props.igvHttpfsUrl}
-                      selectedRecord={state.selectedRecord}
-                      isOpen={state.isViewerOpen}
-                      handlePreviousRecord={this.handlePreviousRecord}
-                      handleNextRecord={this.handleNextRecord}
-                      handleClose={this.handleCloseViewer} />
+        <div className="examine-header">
```
ihodes: How about factoring everything in `examine-header` into its own React component. Additionally, can this div have both `row` and `examine-header` classes? I like avoiding too much nesting if possible/reasonable.

cycledash/static/js/examine/components/CommentBox.js
```diff
@@ -378,26 +380,28 @@ var VCFCommentEditor = React.createClass({
...
-               placeholder='Enter your name here'
-               onChange={this.handleAuthorChange} />
-        <textarea className='form-control comment-textarea'
-                  value={this.state.newCommentText}
-                  placeholder={this.props.placeHolder}
-                  onChange={this.handleTextChange}
-                  ref='textArea' />
-        <div className='save-buttons'>
-          {buttons}
+      <div className='row'>
```
ihodes: Can `row` and `comment-edit-container` be on the same `div`?

cycledash/static/js/examine/components/VCFValidation.js
```diff
@@ -19,11 +19,14 @@ var VCFValidation = React.createClass({
         .map(v => <option value={v.id} key={v.uri}>{v.uri}</option>);
     return (
       <div id='vcf-validations'>
-        Compare to:
-        <select onChange={this.handleSelect}
-                value={this.props.compareToVcfId}>
-          {vcfs}
-        </select>
+        <dl className='dl-horizontal'>
```
ihodes: Do we need this wrapped in the `div`, or can we just add `vcf-validations` to the `dl`?

cycledash/static/js/runs/components/forms.js
```diff
@@ -131,21 +145,24 @@ var NewProjectForm = React.createClass({
...
-        </h2>
-        <h3>New Project</h3>
-
-        <TextInput label='Project Name:' name='name' placeholder='PT01234' required={true} />
-        <div className='form-group project-form-notes'>
-          <label>Notes</label>
-          <textarea className='form-control' rows='3' name='notes'
-                    placeholder='Notes etc.'></textarea>
+      <form method='POST' action='/api/projects' className='add-form' ref='projectForm'>
+        <div className='row'>
```
ihodes: Can this `row` go on the form? (Same with the other forms with a nested `div.row`).

cycledash/static/scss/examine.scss
```diff
@@ -12,39 +12,22 @@ a, a:hover {
...
-  font-size: 16px;
-  line-height: 1;
-  margin: 8px;
-}
 
-/* VCFTable */
+/*------[ VCF Table ]------*/
 .vcf-table {
-  padding-top: 20px;
+  @extend .table, .table-striped, .table-condensed;
```
ihodes: Is there a reason to `@extend` here rather than just add these classes to the `.vcf-table`? 

As an aside: when you `@extend` something, does that mean the CSS from that class is pulled into this file?

ihodes:
Looking good—a few comments, & I'd love to see this this in action/see the pdiff screenshots.

cycledash/static/js/runs/components/forms.js
jaclynperrone: Good call! I removed that extra div and added the `.add-form-input-half` class to the enclosing `.form-group` instead. It's way cleaner to apply the Bootstrap grid to a wrapper div, since you need both the input and its label to sit inside the grid. And Bootstrap has some styling that goes along with their `.form-control` input class that would be messy to override. Will push soon!

cycledash/static/js/examine/components/ExaminePage.js
jaclynperrone: I hear you on the nested. It's all for the Bootstrap grid, though. `.examine-header` is @extending Bootstrap's `.container`, class which consists of some `.row`'s that contain columns.

I'll give the refactoring a shot! But will probably push up these other changes first, since it may take me a while to figure out

cycledash/static/js/examine/components/CommentBox.js
jaclynperrone: Same explanation as above. It's for the Bootstrap grid. But in time this will probably be redesigned, which would change the markup anyway.

cycledash/static/js/examine/components/VCFValidation.js
jaclynperrone: Done!

cycledash/static/js/runs/components/forms.js
jaclynperrone: Same Bootstrap grid explanation as above.

cycledash/static/scss/examine.scss
jaclynperrone: It's a matter of preference/style. So, I like extending Bootstrap classes because it lends to cleaner markup and helps give you more context/control in your stylesheet. It also lets you keep the classes in your markup semantic. For example, you can just name an element what it is (ie `.vcf-table`) and leave the rest to the stylesheet. Also, changing the styling of this object means updating the css, and not fiddling with the markup itself.

For the aside, when you extend class `.table` into `.vcf-table`, `.vcf-table` is now added into the css wherever the `.table` class is being defined. The styles aren't repeated (if that's what you meant?):
screen shot 2015-05-22 at 12 29 45 pm

ihodes: Cool, that seems like a good way to do it :)

cycledash/static/js/examine/components/ExaminePage.js
ihodes: Ah, too bad—dislike non-semantic markup, but if we're saying a refactor/restyling is at hand, this seems reasonable for now :)

ihodes:
LGTM, merge away!

Issue #624, opened 2015-05-20
Show subobjects in JSON responses


ihodes:
e.g. instead of `project_id: <id>` on a run, `project: {id: <id>, name: ..., notes: ...}`

ihodes:
This is required for the `run` information provided on the `/examine` page, removing it regressed https://github.com/hammerlab/cycledash/pull/591

danvk:
This shouldn't be closed until there's a regression test for #630

ihodes:
Not sure this is something we really need/want to do in the API.

PR #630, opened 2015-05-22
Attach tumor/normal bam to the run object being passed to the examine page

ihodes:
Fixes 591 regression caused in the big API change

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/630)


danvk:
We discussed this is Slack. PRs that fix bugs (especially regressions) should always include a test. I'll merge for now, but I've left a note in #624 to remind us to write one.

Issue #615, opened 2015-05-18
Misparsed VCF REF→ALT column


danvk:
http://cycledash.demeter.hpc.mssm.edu/runs/86/examine?query=ORDER+BY+contig%2C+position

`C	→	C[11:128639[`

![slack-imgs com](https://cloud.githubusercontent.com/assets/98301/7686683/a1cda600-fd65-11e4-8a8a-b60d3284b0fd.png)

ihodes:
This is correctly parsed; this is one way that VCFs encode structural variants. 

cf. https://samtools.github.io/hts-specs/VCFv4.2.pdf

cf. http://d.pr/i/yiwN

Issue #559, opened 2015-03-26
PyVCF failing to parse Strelka VCF


ihodes:
cf. http://cycledash.demeter.hpc.mssm.edu/tasks/249

```
Traceback (most recent call last):
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 437, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/cycledash/cycledash/workers/genotype_extractor.py", line 49, in extract
    temporary_dir=TEMPORARY_DIR)
  File "/home/cycledash/cycledash/common/relational_vcf.py", line 169, in insert_genotypes_with_copy
    filename = vcf_to_csv(vcfreader, table_cols, None, **kwargs)
  File "/home/cycledash/cycledash/common/relational_vcf.py", line 128, in vcf_to_csv
    relations = records_to_relations(vcfdata, columns, **kwargs)
  File "/home/cycledash/cycledash/common/relational_vcf.py", line 101, in records_to_relations
    for record in records:
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/vcf/parser.py", line 567, in next
    samples = self._parse_samples(row[9:], fmt, record)
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/vcf/parser.py", line 438, in _parse_samples
    self.samples, samples, samp_fmt, samp_fmt._types, samp_fmt._nums, site)
  File "cparse.pyx", line 54, in vcf.cparse.parse_samples (vcf/cparse.c:1512)
ValueError: could not convert string to float: 
```

ihodes:
According to @arahuja, Strelka also doesn't validate using https://github.com/EBIvariation/vcf-validator

armish:
Turns out strelka doesn't provide data on all the samples it lists in the header and instead creates empty columns for these additional samples. This is problematic to parse, causing PyVCF to fail (as it expects non-empty columns with the proposed format). Removing columns corresponding to `NORMAL.variant2` and `TUMOR.variant2` resolves this problem:

```bash
$ cut -f1-10,12 -d$'\t' strelka.vcf > fixed_strelka.vcf
```

PR #641, opened 2015-05-28
Write test of the genotype extractor

ihodes:
Tests that the extracted genotypes in the database contain the same
information as the VCF records read with pyVCF.

This is the foundation for my plan to test a bunch of different VCF types (e.g. from Somatic Sniper, Guac, VarScan, Mutect) to make sure pyVCF and everything work with them as expected. Cf. issue like https://github.com/hammerlab/cycledash/issues/615 and https://github.com/hammerlab/cycledash/issues/559. Filed issue https://github.com/hammerlab/cycledash/issues/642 to track this.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/641)

### DIFF
--- a/tests/python/test_genotype_extractor_worker.py
+++ b/tests/python/test_genotype_extractor_worker.py
@@ -1,0 +1,132 @@
+import ast
+import mock
+import nose
+import nose.tools as asserts
+import json
+import vcf as pyvcf
+
+from cycledash import app, db
+from common.helpers import tables, find
+
+from test_projects_api import create_project_with_name
+from test_bams_api import create_bam_with_name
+from test_runs_api import create_run_with_uri
+
+from workers.genotype_extractor import _extract
+
+
+class TestGenotypeExtractorWorker(object):
+    def setUp(self):
+        self.project = create_project_with_name('my project')
+        self.run = create_run_with_uri(self.project['id'], '/tests/data/snv.vcf')
+
+    def tearDown(self):
+        with tables(db.engine, 'projects', 'vcfs', 'genotypes') as (con, projects, runs, genotypes):
+            genotypes.delete().execute()
+            runs.delete().execute()
+            projects.delete().execute()
+
+
+    def test_extraction(self):
+        _extract(self.run['id'])
+        with tables(db.engine, 'vcfs', 'genotypes') as (con, vcfs, genotypes):
+            vcf = vcfs.select(vcfs.c.id == self.run['id']).execute().fetchone()
+            variants = genotypes.select(
+                genotypes.c.vcf_id == self.run['id']).execute().fetchall()
+            variants = [dict(v) for v in variants]
+        expected_extant_columns = set(["sample:GT", "sample:RD", "sample:FREQ",
+                                       "info:SSC", "sample:DP", "sample:DP4",
+                                       "info:DP", "info:SS", "info:GPV", "sample:AD",
+                                       "info:SPV"])
+
+        extant_columns = set(ast.literal_eval(vcf['extant_columns']))
+        asserts.eq_(extant_columns, expected_extant_columns)
+        # 10 variants x 2 for the two samples in this file, NORMAL & TUMOR
+        asserts.eq_(vcf['genotype_count'], 20)
+        asserts.eq_(vcf['genotype_count'], len(variants))
+
+        with open('tests/data/snv.vcf', 'r') as vcffile:
+            vcflines = vcffile.readlines()
+        expected_vcf_header = ''.join([l for l in vcflines
+                                       if l.startswith('#')])
+        asserts.eq_(vcf['vcf_header']+'\n', unicode(expected_vcf_header))
+
+        with open('tests/data/snv.vcf') as vcffile:
+            vcfreader = pyvcf.Reader(vcffile)
+            expected_pyvcf_variants = [r for r in vcfreader]
+
+        # Make sure we have the same (chromosome, position, ref, alt) variants
+        assert_keys_in_variants(expected_pyvcf_variants, variants)
+
+        # Make sure we've got the INFO fields we expect
+        pyvcf_getters = [lambda v: v.INFO['DP'],
+                         lambda v: v.INFO['SS'],
+                         lambda v: v.INFO['SSC'],
+                         lambda v: v.INFO['GPV'],
+                         lambda v: v.INFO['SPV']]
+        db_variant_getters = [lambda v: int(v['info:DP']),
+                              lambda v: v['info:SS'],
+                              lambda v: v['info:SSC'],
+                              lambda v: float(v['info:GPV']),
+                              lambda v: float(v['info:SPV'])]
+        assert_keys_in_variants(expected_pyvcf_variants, variants,
+                                expected_getters=pyvcf_getters,
+                                actual_getters=db_variant_getters)
+
+        # Make sure we've got the samples we expect (namely, TUMOR & NORMAL)
+        asserts.eq_(set(vcfreader.samples),
+                    set([v['sample_name'] for v in variants]))
+
+        # Make sure all the NORMAL & TUMOR samples have the same fields with the same values
+        def ensure_samples_equivalent(sample_name):
+            def get_sample_data(attr):
+                def get_from_variant(variant):
+                    sample = find(variant.samples, lambda x: x.sample == sample_name)
+                    return getattr(sample.data, attr)
+                return get_from_variant
+            pyvcf_getters = [get_sample_data('GT'),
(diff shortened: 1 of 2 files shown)


tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+from workers.genotype_extractor import _extract
+
+
+class TestGenotypeExtractorWorker(object):
+    def setUp(self):
+        self.project = create_project_with_name('my project')
+        self.run = create_run_with_uri(self.project['id'], '/tests/data/snv.vcf')
+
+    def tearDown(self):
+        with tables(db.engine, 'projects', 'vcfs', 'genotypes') as (con, projects, runs, genotypes):
```
tavinathanson: Since you're doing this in multiple tests, maybe worth a helper to `delete_tables(genotypes, runs, projects)`?

workers/genotype_extractor.py
```diff
@@ -21,6 +21,11 @@
 
 @worker.task(bind=True)
 def extract(self, vcf_id):
+    register_running_task(self, vcf_id)
+    return _extract(vcf_id)
```
tavinathanson: What does `register_running_task` do that you didn't want to do in a test?

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+        with tables(db.engine, 'vcfs', 'genotypes') as (con, vcfs, genotypes):
+            vcf = vcfs.select(vcfs.c.id == self.run['id']).execute().fetchone()
+            variants = genotypes.select(
+                genotypes.c.vcf_id == self.run['id']).execute().fetchall()
+            variants = [dict(v) for v in variants]
+        expected_extant_columns = set(["sample:GT", "sample:RD", "sample:FREQ",
+                                       "info:SSC", "sample:DP", "sample:DP4",
+                                       "info:DP", "info:SS", "info:GPV", "sample:AD",
+                                       "info:SPV"])
+        expected_vcf_header = """\
```
tavinathanson: I feel like this blob makes the test harder to read. How about a data file?

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+                                 lambda v: int(v['sample:AD']),
+                                 lambda v: v['sample:FREQ'],
+                                 lambda v: tuple(ast.literal_eval(v['sample:DP4']))]
+            tumor_variants = [v for v in variants
+                              if v['sample_name'] == sample_name]
+            assert_keys_in_variants(expected_variants, tumor_variants,
+                                    expected_extractors=expected_extractors,
+                                    actual_extractors=actual_extractors)
+
+        _ensure_samples_equivalent('NORMAL')
```
tavinathanson: Unclear about how to chose the `_` prefix for this method?

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+        with tables(db.engine, 'projects', 'vcfs', 'genotypes') as (con, projects, runs, genotypes):
+            genotypes.delete().execute()
+            runs.delete().execute()
+            projects.delete().execute()
+
+
+    def test_extraction(self):
+        _extract(self.run['id'])
+        with tables(db.engine, 'vcfs', 'genotypes') as (con, vcfs, genotypes):
+            vcf = vcfs.select(vcfs.c.id == self.run['id']).execute().fetchone()
```
tavinathanson: So, what are you doing with `vcf`?

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+
+        _ensure_samples_equivalent('NORMAL')
+        _ensure_samples_equivalent('TUMOR')
+
+
+
+def assert_keys_in_variants(
+        expected_variants, actual_variants,
+        expected_extractors=[], actual_extractors=[]):
+    """Asserts that the variants in in expected and actual match each other on
```
tavinathanson: "in in"

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+    chromosome, position, reference, alternates, and the result of applying the
+    functions in expected_extractors and actual_extractors to expected_variants
+    and actual_variants, respectively.
+    """
+    def extract_expected_keys(v):
+        base = [v.CHROM, v.POS, v.REF, ''.join(map(str, v.ALT))]
```
tavinathanson: Wow, it took me a long time to understand why you were using `v.CHROM` here `v['contig']` down there. Maybe renaming `v` here to `pyvcf_variant` and `v` below to `db_variant`, or something to that effect?

tests/python/test_genotype_extractor_worker.py
```diff
@@ -0,0 +1,144 @@
...
+        asserts.eq_(vcf['vcf_header'], expected_vcf_header)
+
+        vcfreader = pyvcf.Reader(open('tests/data/snv.vcf'))
+        expected_variants = [r for r in vcfreader]
+
+        # Make sure we have the same (chromosome, position, ref, alt) variants
+        assert_keys_in_variants(expected_variants, variants)
+
+        # Make sure we've got the INFO fields we expect
+        expected_extractors = [lambda v: v.INFO['DP'],
```
tavinathanson: I got confused and thought you were literally trying to compare extractors (the "expected" vs. "actual" extractor). Maybe call this `pyvcf_extractors` or something?

tavinathanson: Better yet, avoid using `extractor` altogether since you're testing an `extractor`. `pyvcf_value_pickers`?

tavinathanson:
Back to you @ihodes

tests/python/test_genotype_extractor_worker.py
ihodes: done in just-merged PR, will do this when this is merged in

workers/genotype_extractor.py
ihodes: It registered the task with the database; it required that we run the worker queue, so I don't want to test that.

tests/python/test_genotype_extractor_worker.py
ihodes: :+1:

workers/genotype_extractor.py
tavinathanson: Can you just mock it out or something rather than creating confusing (IMO) `_` methods?

tests/python/test_genotype_extractor_worker.py
ihodes: I'll nix it—as discussed offline

tests/python/test_genotype_extractor_worker.py
ihodes: I use it below to make sure the database matches up with the VCF file itself.

tests/python/test_genotype_extractor_worker.py
ihodes: :+1:

workers/genotype_extractor.py
ihodes: I can do that, but I'd still rather have an `_extract` method: in this case I think the underscore is warranted; the primary export of this module is the worker task, but the worker task is doing something somewhat different than `_extract` alone (namely registering itself, plus everything the decorator is doing [wrapping it in a class specific to Celery, I believe?]).

ihodes: I think it's pretty standard in Python to do stuff like this.

tavinathanson: Fair enough; I've seen `_` a lot, but I still find it confusing :)

tests/python/test_genotype_extractor_worker.py
tavinathanson: My bad; ctrl-F had apparently failed me

workers/genotype_extractor.py
ihodes: I just think of it as "don't rely on this method; it's not part of the API, but internal to the module".

tests/python/test_genotype_extractor_worker.py
ihodes: I know how that goes.

tests/python/test_genotype_extractor_worker.py
ihodes: I switched to `getters`—how's that sound?

ihodes:
Thanks for the review—this wasn't a fun one, I know—back to you!

PR #678, opened 2015-06-02
Add tests for genotypes api, fix contig ordering

### LINKED ISSUE #636: Test the cycledash.genotypes module

ihodes:
fix https://github.com/hammerlab/cycledash/issues/407 & adds regression test for it + lots of other genotypes tests (ticking off some todos in #636, and setting up easy template for the rest)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/678)

### DIFF
--- a/common/helpers.py
+++ b/common/helpers.py
@@ -66,6 +66,11 @@ def find(iterable, pred):
             return x
 
 
+def pick(obj, *keys):
+    """Return an object with only the key/vals with key name in `keys`."""
+    return {key: obj[key] for key in keys}
+
+
 def to_epoch(dt):
     """Return the epoch time representation of a datetime object, `dt`."""
     epoch = datetime(1970, 1, 1, tzinfo=dt.tzinfo)
--- a/cycledash/genotypes.py
+++ b/cycledash/genotypes.py
@@ -9,7 +9,7 @@ from flask.ext.restful import Resource
 from sqlalchemy import (select, func, types, cast, join, outerjoin, asc, desc,
                         and_, Integer, Float, String)
 from sqlalchemy.sql import text
-from sqlalchemy.sql.expression import label, column
+from sqlalchemy.sql.expression import label, column, case, literal
 from sqlalchemy.sql.functions import coalesce
 import vcf as pyvcf
 from plone.memoize import forever
@@ -69,7 +69,7 @@ def contigs(vcf_id):
     return [contig for (contig,) in results]
 
 
-def get(vcf_id, query, with_stats=True):
+def get(run_id, query, with_stats=True):
     """Return a list of genotypes in a vcf conforming to the given query, as
     well as a dict of stats calculated on them.
 
@@ -92,16 +92,16 @@ def get(vcf_id, query, with_stats=True):
      limit: 250
     }
     """
-    query = _annotate_query_with_types(query, spec(vcf_id))
-    compare_to_vcf_id = query.get('compareToVcfId')
+    query = _annotate_query_with_types(query, spec(run_id))
+    compare_to_run_id = query.get('compareToVcfId')
     with tables(db.engine, 'genotypes') as (con, g):
-        if compare_to_vcf_id:
+        if compare_to_run_id:
             # We consider a genotype validated if a truth genotype exists at its
             # location (contig/position) with the same ref/alts.  This isn't
             # entirely accurate: for example, it handles SVs very poorly.
             gt = g.alias()
             joined_q = outerjoin(g, gt, and_(
-                gt.c.vcf_id == compare_to_vcf_id,
+                gt.c.vcf_id == compare_to_run_id,
                 g.c.contig == gt.c.contig,
                 g.c.position == gt.c.position,
                 g.c.reference == gt.c.reference,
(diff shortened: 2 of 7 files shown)


common/helpers.py
```diff
@@ -66,6 +66,11 @@ def find(iterable, pred):
             return x
 
 
+def pluck(obj, *keys):
```
tavinathanson: http://funcy.readthedocs.org/en/latest/colls.html#pluck?

cycledash/genotypes.py
```diff
@@ -92,16 +92,16 @@ def get(vcf_id, query, with_stats=True):
      limit: 250
     }
     """
-    query = _annotate_query_with_types(query, spec(vcf_id))
-    compare_to_vcf_id = query.get('compareToVcfId')
+    query = _annotate_query_with_types(query, spec(run_id))
+    compare_to_run_id = query.get('compareToVcfId')
```
tavinathanson: Confused about where we're using `run` and where we're using `vcf`; this line is a good example of both

cycledash/genotypes.py
```diff
@@ -120,9 +120,9 @@ def get(vcf_id, query, with_stats=True):
 
         q = _add_ordering(q, g, 'String', 'contig', 'asc')
         q = _add_ordering(q, g, 'Integer', 'position', 'asc')
-
+        #print str(q.compile(compile_kwargs={"literal_binds": True}))
```
tavinathanson: Get rid of this commented line?

cycledash/genotypes.py
ihodes: Yeah…this is part of a slow migration from VCF to run. I've just been changing it where I touch the code. I'll change `compareToVcfId` to `compareToRunIf`.

tests/python/test.py
```diff
@@ -14,7 +15,7 @@ def test_underscorize():
         "CamelBI1GCasesRun1": "camel_bi1g_cases_run1"
     }
     for example, expected in camel_to_underscored_expectations.items():
-        assert underscorize(example) == expected
+        asserts.eq_(underscorize(example), expected)
```
tavinathanson: Nit: maybe just `from nose.tools import eq_`?

common/helpers.py
ihodes: Ah, I meant to call it pick. cf. http://underscorejs.org/#pick

Will fix. 

Might be worth bringing in funcy at some point?

cycledash/genotypes.py
ihodes: Done

tests/python/test.py
```diff
@@ -51,6 +52,13 @@ def test_order():
...
+    asserts.eq_(o1, [99, 22, 132])
+    asserts.eq_(o2, [{'thing': 'toga'}, {'thing': 'bathrobe'}])
+    asserts.eq_(o3, [1, -2, 2, 4, 4])
+
+
+@asserts.raises(KeyError)
+def test_pluck():
+    asserts.eq_(pluck({'a': 1, 'b': 2, 'c': 'abc'}, 'a', 'c'), {'a': 1, 'c': 'abc'})
+    asserts.eq_(pluck({'a': 1, 'b': 2, 'c': 'abc'}), {})
+    pluck({'a': 'oh no'}, 'b')  # This rasies a KeyError
```
tavinathanson: "raises"

tests/python/test.py
ihodes: I thought about that, but I like that is still reads as "assert equal", though—what think you?

tests/python/test.py
ihodes: Yeesh, done.

tests/python/test_genotypes_api.py
```diff
@@ -0,0 +1,206 @@
+import mock
+import nose
+import nose.tools as asserts
+import json
+
+from cycledash import db
```
tavinathanson: Do we have a story around line breaks between imports? Personally, I've been generally doing 1 group of imports for internal and (with an extra line break in between the two groups) 1 group of imports for external

ihodes: Hmm, not a great story AFAIK. I think your way is to the style guide, and I'll change this to be "external \n, cycledash-specific \n, in-current-test-directory" -- does that seem reasonable?

common/helpers.py
tavinathanson: Not sure how stable/etc. `funcy` is, but I guess that's what I'd do if it's solid (or if there's something else that's solid).

cycledash/genotypes.py
tavinathanson: Gotcha

tests/python/test.py
tavinathanson: Fair enough

tests/python/test_genotypes_api.py
tavinathanson: :+1:

tavinathanson:
Fly, good Fleance, fly, fly, fly (and merge)

cycledash/genotypes.py
ihodes: Hmm, maybe holding off on this for now is okay? This is from the Javascript side of the application; seems to touch a lot of code… I'm also okay reverting this change if you think that'd be preferable?

tavinathanson: Fine with me

PR #702, opened 2015-06-09
Pass the new &format= param to IGV

### LINKED ISSUE #574: IGV integration broken

danvk:
Fixes #574 

I verified that this results in a correctly displayed VCF track for run 270, which has a VCF file ending in `part-r-00000`, rather than `.vcf`.

cc @arahuja

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/702)

### DIFF
--- a/cycledash/static/js/examine/utils.js
+++ b/cycledash/static/js/examine/utils.js
@@ -36,16 +36,17 @@ function makeIGVLink(run, igvHttpfsUrl) {
   }
 
   var nameFilePairs = [
-      ['Run', run.uri],
-      ['Normal', run.normal_bam && run.normal_bam.uri],
-      ['Tumor', run.tumor_bam && run.tumor_bam.uri]
-  ].filter(x => x[1]);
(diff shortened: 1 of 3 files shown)


danvk:
I dropped the `maxlen` setting from `.jshintrc`. It was previously set to 200. I'm not sure what the value of such a large setting is, and I'd prefer not to split expected URLs onto multiple lines to satisfy the linter.

cycledash/static/js/examine/utils.js
```diff
@@ -36,16 +36,17 @@ function makeIGVLink(run, igvHttpfsUrl) {
   }
 
   var nameFilePairs = [
```
ihodes: This name doesn't match the value anymore; it could be clearer to have an array called 'tracks' of objects with keys like `name`, `type`, and `uri`.

danvk: Sure, done.

PR #708, opened 2015-06-09
Use livereactload to reload new JS without refreshing the page

ihodes:
Like magic.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/708)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,6 +2,7 @@ var _ = require('underscore'),
     browserify = require('browserify'),
     gulp = require('gulp'),
     livereload = require('gulp-livereload'),
+    livereactload = require('livereactload'),
     peg = require('gulp-peg'),
     reactify = require('reactify'),
     source = require('vinyl-source-stream'),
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -39,7 +40,10 @@ var REACT_OPTS = {es6: true},
 // Runs JSX, ES6 transforms, browserify, adds sourcemaps, and notifies the
 // livereload server that JS has changed.
 gulp.task('js', function() {
-  var bundler = watchify(browserify(BROWSERIFY_OPTS));
+  livereactload.monitor(PATHS.dest + 'bundled.js')
+
+  var bundler = watchify(browserify(_.extend({transform: [livereactload]}, BROWSERIFY_OPTS)));
+  rebundle();
```
danvk: this looks a bit odd — you call `rebundle()` before it's defined, then immediately call it again?

ihodes: It's callable at this point (you can call functions earlier in the file from where they're defined; seems quite common in JS (in fact, this is a copy-pasta error). Killed the extra rebundle() call though :)

danvk:
Just because you can doesn't mean you should :)

LGTM, feel free to merge on green.

PR #714, opened 2015-06-10
Better exception formatting.

ihodes:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/714)

### DIFF
--- a/cycledash/__init__.py
+++ b/cycledash/__init__.py
@@ -2,6 +2,8 @@ from flask import Flask, jsonify, request
 from flask_sqlalchemy import SQLAlchemy
 from flask.ext import restful
 import humanize
+import logging
+import sys
 
 
 def initialize_application():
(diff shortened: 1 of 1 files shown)


cycledash/__init__.py
```diff
@@ -15,19 +17,15 @@ def initialize_application():
...
-        app.logger.exception(exception)
-        return 'Internal Server Error', 500
-
-    import logging
-    import sys
-
     stdout_handler = logging.StreamHandler(sys.stdout)
     stdout_handler.setLevel(logging.INFO)
-    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    formatter = logging.Formatter("""\
```
danvk: is this a literal `\`?

danvk:
Worth noting that we lose our ability to grep logs by going to a multiline format, but that's not something we've ever done in the past.

ihodes:
Our logs have actually always been in a multi line format! This just adds one more newline (+ some more info) to make it easier to read & grep.

cycledash/__init__.py
ihodes: No, it escapes the newline,

danvk: weird, did not know this trick.

PR #721, opened 2015-06-12
Full page pileup

### LINKED ISSUE #718: Show pileup view as a full-page when launched
From @danvk:
When you click "Open Pileup Viewer", it will show a full-page pileup viewer, with some information about the selected row on the top. 

When you close the full-page pileup viewer (by hitting the "x" icon), the variant you were looking at will remain selected. This will facilitate the Inspect at a variant → Comment on that variant workflow.

Full page by default will work well on laptop screens. It's unclear how well it will work on our giganto cinema displays.

---
JP: This should address the bug with the height of the panel. Right now it's tall and is getting cut off in the browser. And also the bug where scrolling to the top of the pileup view causes the table in the background to scroll, too.

danvk:
Fixes #718 

This also fixes the "mousewheel in the pileup track scrolls the variant list" issue. Mousewheeling elsewhere in the variant inspector still does this, but the situation is at least improved!

The full-page pileup view includes a one-row `VCFTable`. In setting this up, I found a few unused props which I removed.

This uses CSS flexbox to size the tracks in the pileup view. This works so well that I might make it the default in `pileup.css`.

full page pileup

@jaclynperrone

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/721)


ihodes:
LGTM, two minor issues:

1) http://d.pr/i/1edTw
2) Some way to get rid of the scrollbar on the main page would be ideal; can't we just not render the main VCFTable if pileup is open?

danvk:
Good suggestion about not rendering the VCFTable when the pileup viewer is open. I reworked `ExaminePage.js` to show only the pileup or the rest of the examine page, never both. This fixes all three of your minor issues. It also lets me kill the mousewheel hack.

Lemme know what you think.

ihodes:
Awesome, LGTM!

PR #726, opened 2015-06-12
Use the newest version of pileup with require support

armish:
- upgrade [pileup.js](https://github.com/hammerlab/pileup.js) to `v0.1.4` and access the module with `require('pileup')` instead of depending on global variables
- when bundling the source, _reactify_ globally so that dependencies also get transformed (the new [pileup.js](https://github.com/hammerlab/pileup.js) needs this until we resolve this issue: https://github.com/hammerlab/pileup.js/issues/188)
- upgrade the [uglifyify](https://www.npmjs.com/package/uglifyify) as the older version was not able to minify React properly causing conflicts across pileup.js and Cycledash instances

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/726)

### DIFF
--- a/cycledash/static/js/examine/components/PileupViewer.js
+++ b/cycledash/static/js/examine/components/PileupViewer.js
@@ -2,10 +2,9 @@
 
 var React = require('react'),
     _ = require('underscore'),
+    pileup = require('pileup'),
     $ = require('jquery'),
     VCFTable = require('./VCFTable');
-// Someday: pileup = require('pileup');
-
 
 // Indicator value that the BAI chunks are still loading.
 var CHUNKS_LOADING = 'loading',
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -29,7 +29,7 @@ var PATHS = {
   ]  // polyfills
 };
 
-var REACT_OPTS = {es6: true},
+var REACT_OPTS = {es6: true, global: true},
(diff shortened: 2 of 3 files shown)


ihodes:
Nice!

armish:
It took me a while to fix the React clashing problem, but turns out it was just `uglifyify`'s old implementation that was behaving badly :\ 

Have a nice weekend and safe travels, @ihodes!

danvk:
@armish is this good to merge? Or do you want to find a different (flow compatible) way to resolve the browserify issues?

armish:
@danvk haven't resolved the flow issue yet. Other than the flow issue, this works fine; but, feel free to hold off from merging this before we completely resolve issues on pileup.js side.

danvk:
Let's go ahead and merge this—it should be the same after we fix things up on the pileup side.

Could you check what the size of `cycledash/static/js/dist/bundled.js` is before & after this change? I'm hoping that this results in Cycledash and pileup.js sharing copies of React, D3 and underscore.

armish:
unfortunately this changeset does not make them share those libraries yet, as pileup.js is still included as a bundled file with all the libraries in it. The size of the `bundled.js` was 620K before the changes and it is now ~1M. The new uglifyify seems to be doing a worse job :\

PR #745, opened 2015-06-17
Use glyphicons

jaclynperrone:
This commit does two things:
- Moves the glyphicon font files from `lib/bootstrap/fonts` into the `static/fonts` folder via gulp.task. This corrects the broken relative file path for the glyphicon font
  
- Replaces (and deletes!) svgs with glyphicons

Glyphicons are more flexible than svgs, in that you can easily manipulate them with css. And now you don't have to load extra image files. Yeah!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/745)


danvk:
Looks good!

PR #750, opened 2015-06-17
Add a varcode-powered worker for annotating VCF files

armish:
This new worker annotates each variant in a VCF file with the top priority effect properties:
- Gene name (Official gene symbol)
- Transcript ID (Ensembl ID for the transcript)
- Protein-level change (*e.g.* V600E)
- Type of the variant (*.e.g* Substitution, Splice Site, Intronic, ...)

This worker is run when user submits a new VCF file (*i.e.* when a new **run** is created)
screen shot 2015-06-17 at 5 50 15 pm

Once the worker completes its annotation, new information on variants are shown under `Annotations` Column Group with the *Varcode* prefix:
screen shot 2015-06-17 at 5 49 51 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/750)


cycledash/genotypes.py
```diff
@@ -397,14 +420,6 @@ def _header_spec(vcf_header_text, extant_cols):
                 'Names of genes that overlap with this variant\'s '
                 'starting position, derived from Ensembl Release 75.'))
 
-    # Remove empty supercolumns
```
ihodes: Did something change to obsolete this?

workers/gene_annotator.py
```diff
@@ -70,6 +70,7 @@ def annotate(self, vcf_id):
 
         # We've added annotations:gene_names, so update the columns to display.
         update_extant_columns(metadata, con, vcf_id)
+        return(vcf_id)
```
ihodes: No need for `()` in a return

ihodes:
Awesome to have this coming in! Thanks for the work, @armish. 

I do think we can drop the gene_annotator; there's no reason to have both now, AFAIK. With that, I'd drop the `varcode_` prefix to the column names as well. 

Could you also add the migration code required to get the database up to date with the  new schema in the PR text? We do this to make deployment slightly easier. It's definitely not the right solution… but it's something :)

cycledash/genotypes.py
```diff
@@ -383,8 +383,31 @@ def _header_spec(vcf_header_text, extant_cols):
         path=['sample_name']) # This path is not the default super -> sub column
 
     # Add Cycledash-derived columns
-    column_name = 'annotations:gene_names'
-    if column_name in extant_cols:
+    _add_extant_column_to_spec(extant_cols, 'annotations:gene_names', res, 'String', 1,
```
ihodes: I can't tell if it's just some funky GitHub formatting, but some of these lines around here may be longer than 80 charts; mind chopping them down to size?

cycledash/genotypes.py
```diff
@@ -383,8 +383,31 @@ def _header_spec(vcf_header_text, extant_cols):
...
+            'derived from Varcode project.'))
+
+    # Remove empty supercolumns
+    for key, val in res.iteritems():
+        if not val.keys():
+            del res[key]
+
+    return res
+
+def _add_extant_column_to_spec(extant_cols, column_name, res, column_type, num, description):
```
ihodes: I'd add a \n + some spaces between `spec(` and `extant_cols` to get this below 80 chars (ignore these comments if it's under 80!)

ihodes: Ah, any reason not to use the below `_add_column_to_spec`?

workers/varcode_annotator.py
```diff
@@ -0,0 +1,133 @@
...
+        return  # An error must have occurred earlier.
+    register_running_task(self, vcf_id)
+
+    EnsemblRelease(config.ENSEMBL_RELEASE).install()  # Only runs the first time for this release.
+
+    engine = sqlalchemy.create_engine(DATABASE_URI)
+    with tables(engine, 'genotypes') as (con, genotypes):
+        metadata = sqlalchemy.MetaData(bind=con)
+        metadata.reflect()
+        gene_names = get_varcode_annotations(genotypes, vcf_id, config.ENSEMBL_RELEASE)
```
ihodes: Maybe `gene_names` should be `annotations` or something? In general, you may consider dropping the `varcode_` prefix you have in some places. I don't think it's necessary (though I can see the argument & use for having it in the module name, as this module just annotates the loci with all the annotations Varcode generates).

workers/varcode_annotator.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # This will give us a single, yet relevant effect
+        best_effect = variant.effects().top_priority_effect()
+        gene_name = best_effect.gene_name
+        transcript = best_effect.transcript_id
+        notation = best_effect.short_description
+        effect_type = type(best_effect).__name__
+        # Make it human readable
+        effect_type = re.sub("([a-z])([A-Z])","\g<1> \g<2>", effect_type)
+        varcode_annotations.append([contig, position, reference, alternates,
```
ihodes: Indentation should be pep8-y cf. http://legacy.python.org/dev/peps/pep-0008/#indentation

(Need to line up the hanging indent, in this case)

cycledash/genotypes.py
armish: No, actually not -- this is just an artifact of the diff. These lines are still present a little bit above, but somehow this diff view thinks that they were deleted and added back.

cycledash/genotypes.py
armish: The only reason was to not duplicate this logic for every single one of them:

```
   ...
   if column_name in extant_cols:
        supercolumn, subcolumn = column_name.split(':')
   ...
```
If we are going to assign custom column names to these, it might make sense to simply use `_add_column_to_spec`, IMHO. Any ideas?

workers/gene_annotator.py
armish: ah, old habits :\

workers/varcode_annotator.py
armish: yeah, it makes sense to drop the `varcode_` prefix. The `gene_names` variable name was a result of extensive copy/pasting across annotators. Will definitely fix that.

workers/varcode_annotator.py
armish: Noted -- sorry, didn't know about indentation practices; also thanks for the pointer. These guidelines are really helpful.

cycledash/genotypes.py
armish: Yeah, will definitely do.

armish:
@ihodes: and here is the migration code:

```
ALTER TABLE genotypes
    DROP COLUMN "annotations:gene_names",
    ADD "annotations:gene_name" TEXT,
    ADD "annotations:transcript" TEXT,
    ADD "annotations:effect_notation" TEXT,
    ADD "annotations:effect_type" TEXT;
```

that is, if Travis can give us the green light :)

schema.sql
```diff
@@ -69,7 +69,11 @@ CREATE TABLE genotypes (
        quality TEXT,
 
        -- Cycledash-derived data
-       "annotations:gene_names" TEXT,
+       -- These are from the varcode annotation worker
+       "annotations:gene_name" TEXT,
```
ihodes: If there are multiple genes, do they still get displayed? (If so, this should probably still be "annotations:gene_name**s**")

armish: Nope -- this annotation only picks an effect in a single gene. So we won't be having multiple genes annotations any more.

ihodes: Hmm—is there a way to get that back? That seems like a useful feature /cc @iskandr @tavinathanson

ihodes:
This is looking good! We still need these to show up in our screenshot tests as well, cf. 

>  this will also require updating the test database used for generating screenshots, and generating new screenshots for the examine page. Cf. the tests/pdifftests directory for more info.

schema.sql
iskandr: @armish You only have a single gene if you're picking the top priority annotation for each variant. 

You could do this to get the top effect for each distinct gene:

```
for gene_name, gene_effects in variant.effects().groupby_gene_name().items():
    top_gene_effect = gene_effects.top_priority_effect()
```

but you'll have to then deal with all sorts of weird non-coding gene types.

armish: From a programmatic point of view, there is no reason not to include multiple gene names; but putting my biologist hat on, I think we should only show a single gene to the users. This might be a bold step forward, but I opt for single gene/variant route.

Happy to take this discussion off-line ;)

armish:
phew! It took me a few iterations to get it right, but here we are with the updated screenshots and a successful build \o/

schema.sql
ihodes: I don't have a very string opinion about this, but could you file this in an issue to resolve out of band with this PR?

ihodes:
This looks great! Thanks for putting it together—will be a great addition to Cycledash.

PR #773, opened 2015-07-01
Update DEVELOP.md

ihodes:
Add info about daemonizing SauceConnect

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/773)

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -184,3 +184,13 @@ suite.
 To determine whether there are any pixels that have changed before/after, and to
 generate a perceptual diff that will make it clear where the changes are, you
 can use [webdiff](https://github.com/danvk/webdiff): `git webdiff`.
+
+You may speed up the `update-screenshots.sh` script by daemonizing SauceConnect
+in advance. 
+
+```
+sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY --daemonize
+```
+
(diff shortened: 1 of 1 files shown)


danvk:
Please include the bit about `--wait`, too. You may want to verify with @jaclynperrone that this did actually speed up the update-screenshots command (if it did, we're not seeing any evidence of it).

ihodes:
The --wait flag is seltest specific. I'd rather point to the docs then re-document seltest here. (Additionally, interactive is more useful than wait). 




I'd be surprised if this isn't speeding the script up. Will take a look at it on her machine.

jaclynperrone:
@ihodes That would be super helpful! It's taken about 5 minutes to generate the first 8 screenshots.

danvk:
Mentioning that `--wait` and `sel interactive` are useful & linking off to the docs would be really helpful. Neither Jaclyn nor I knew to try those.

ihodes:
I'll add some notes + pointers, makes sense. 

FTR, the slowness appears to be the tunnel between sauce and our office is slow today; we observed this on the SauceLabs videos (pages just taking a long time to load), and saw that the tests when connecting to Travis is much faster.

ihodes:
@dan back to you (I left out --wait as mentioned, as I wouldn't recommended it).

Issue #710, opened 2015-06-09
Redesign the VCF comparison table and examine page


jaclynperrone:
Some things to consider:
- Make VCF meta data smaller
- Move comparison dropdown near CQL search
- Bring VCF comparison table near CQL search

jaclynperrone:
Closed by #774

jaclynperrone:
Closed by PR #774

PR #774, opened 2015-07-01
Examine page header

jaclynperrone:
Would love some feedback for this one! This is the first pass at redesigning the header on the examine page. This commit does a few things:

- Move CQL filter, VCF comparison and # of records into the table header
- Only the VCF table horizontally scrolls, instead of the entire page
- `RecordsShown` is now its own component (instead of being part of `StatsSummary`). This is to give greater flexibility to the layout.
- Updates pdiff screenshots

Thanks!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/774)


jaclynperrone:
Pulling in @danvk as well for some feedback!

danvk:
There are some clipping issues with the screenshots, but I think the redesign looks great!

ihodes:
Looks great, big improvement! I wouldn't mind the stats taking up less room somehow—they get a ton of space. Maybe file an issue to do this if you have plans to change it more?

tests/pdifftests/test.py
```diff
@@ -3,10 +3,8 @@
 from selenium.webdriver.support.ui import Select
 from selenium.webdriver.common.action_chains import ActionChains
```
ihodes: This is a massive silly nit, so don't worry about changing it, but in general 2 newlines between imports and top-level definitions.

jaclynperrone: Ahhh good to know! Thought I was doing a little light cleaning :)

jaclynperrone:
@danvk @ihodes Thanks for taking a look!!

@ihodes Those stats have been tricky, that's for sure. It makes sense that they live where they live, and I agree that they take up a lot of room. I am going to reduce the padding a little bit to tighten them up. And while I'm at it, I'll fix that `test.py` file as well :)

I think that any further changes to those stats will be handled in another PR. I would love sit together and figure out a friendlier, more straightforward way of presenting that information. I remember we played around with the idea of having a legend that says what the symbols mean after the comparison is done.

jaclynperrone:
Closes #710

PR #777, opened 2015-07-06
Pileup style tweaks in Cycledash.

jaclynperrone:
- Updates css to accommodate pileup 0.2 release
- Replaces previous, next, zoom and close buttons with glyphicons

screen shot 2015-07-06 at 11 54 59 am

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/777)


jaclynperrone:
Looping @danvk in. Would love some feedback from you as well!

danvk:
It's pileup 0.2, not 2.0 :)

ihodes:
This looks splendid, thank you. Do the arrow keys/esc still work? (Did they ever work? I may be making things up.)

cycledash/static/scss/examine.scss
```diff
@@ -549,7 +584,7 @@ a.download-vcf {
 }
 
 #pileup-container .controls {
-  flex: 0 0 35px;  /* fixed height */
+  flex: 0 0 38px;  /* fixed height */
```
danvk: The added height is to accommodate the larger zoom in/out buttons?

danvk:
@ihodes they used to but they don't now because I've punted on a notion of whether the pileup viewer has focus. Feel free to file an issue to add them back.

@jaclynperrone This looks great. One question is whether the big prev/next buttons should be up/down or left/right. You may or may not be moving backward/forward in the genome (if the variant list is sorted by position). You're definitely moving up/down in the list. I could go either way on this, up/down buttons might feel a bit odd in that position.

Otherwise looks great!

ihodes:
@dan https://github.com/hammerlab/cycledash/issues/696 https://github.com/hammerlab/cycledash/issues/778

cycledash/static/scss/examine.scss
jaclynperrone: Yup! But I actually don't need that much height anymore. Gonna push a tiny tweak.

jaclynperrone:
@danvk I would definitely stick with the left/right arrows. I think the up/down would be confusing. I just pushed a tiny update to the control panel height.

@ihodes Thanks for taking a look! Glad you like it :)

danvk:
LGTM

Issue #805, opened 2015-08-03
pileup tracks scroll vertically


danvk:
The location, scale, genome and VCF tracks all scroll by a few pixels. They shouldn't:

![cycledash updated](https://cloud.githubusercontent.com/assets/98301/9048043/186b067e-3a04-11e5-9bec-cf4012de8bc4.png)

There seem to be 5px of mysterious vertical space in the `<div>` that contains each `<svg>` element.

armish:
I didn't know this before, but after some fiddling I was able to isolate the problem into this one:
http://jsfiddle.net/yh7L7wwa/

and learned found [a trick on SO](http://stackoverflow.com/questions/24626908/how-to-get-rid-of-extra-space-below-svg-in-div-element) to get rid of the space there:
http://jsfiddle.net/yh7L7wwa/1/

Looks like we need to give the examine page some CSS love (will do).

armish:
addressed by #811.

PR #811, opened 2015-08-20
Examine page: remove mysterious vertical space within pileup track elements

armish:
See #805 for more context.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/811)

### DIFF
--- a/cycledash/static/scss/examine.scss
+++ b/cycledash/static/scss/examine.scss
@@ -297,7 +297,7 @@ a.download-vcf {
   background-color: $color-bg-hover;
   border: 1px solid $color-border-light;
   min-width: 200px;
-  font-size: 0.75rem; 
+  font-size: 0.75rem;
   color: $color-text;
   z-index: 1000;
   opacity: 100; /* This is needed to override Bootstrap. */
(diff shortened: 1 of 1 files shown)


danvk:
Can you make this change in pileup, rather than cycledash?

armish:
Sure @danvk, I can do that; but I thought we were having this problem because of the custom CSS used in Cycledash. We don't see this behavior in pileup.js demo page, right? Or do you want to just make sure that this never happens when using pileup.js? If so, I will do the changes in pileup.

danvk:
This is the vertical scrolling issue, right? I'm pretty sure that does
happen in the pileup demo, too, but please double check!

armish:
yeah, this addresses the vertical scrolling issue. I wasn't able to replicate this using the latest pileup.js code and the playground example we have there:

screen shot 2015-08-20 at 4 18 30 pm

Having said this, I also don't know and wasn't able to find which CSS property triggers this behavior in Cycledash :\ I will take another look and update/move the PR if I can nail it down.

danvk:
Cool. You may be able to scroll tracks in the demo even if you can't see a
scroll bar (using the mouse wheel)

armish:
@danvk: cannot replicate these mysterious space problem within pileup.js. I am assining this PR to you so that when you are back, we can take a look at it together and decide where to make these changes.

PR #814, opened 2015-08-21
Update screenshots to fix broken CI tests on Travis

armish:
For some reason unknown to me, our CI tests were failing us due to differing screenshots. See [this trivial change](https://github.com/hammerlab/cycledash/pull/812) in the README file and [the corresponding fail report](https://travis-ci.org/hammerlab/cycledash/builds/76526407) from Travis.

A few pages were off had different renderings of button corners (like run pages). I am not sure why these need an update, but might be due to an update on the Sauce Labs side that affected renderings.

This is to bring the screenshots up-to-date and resolve issues with Travis.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/814)


armish:
FYI: I will merge this trivial one in and continue working on the other ones.

scripts/update-screenshots.sh
```diff
@@ -82,9 +82,17 @@ else
     echo "SauceConnect tunnel detected already running, will attempt to use that."
 fi
 
+capabilities=$(python <<EOF
+import json
+capabilities = json.loads(open('$SELTEST_CAPABILITIES').read())
+capabilities['tunnel-identifier'] = '$TRAVIS_JOB_NUMBER'
```
ihodes: ./update-screenshots.sh runs locally, so there won't be a TRAVIS_JOB_NUMBER--I think this file should remain the same.

ihodes:
Sorry I didn't get to this earlier! The ./update-screenshots.sh script needs to be reverted to the way it was before; cf. my comment on it.

I'm not quite sure, but it looks like the issue was sparked this PR was fixed (timestamp issues?), and the screenshots needed to be changed in the end just because of the Chrome version bump to 44?

Issue #738, opened 2015-06-15
Download VCF gives an "Internal Server Error"


arahuja:

ihodes:
Mind linking a particular run?

arahuja:
http://cycledash.demeter.hpc.mssm.edu/runs/327/examine?query=sample_name+%3D+TUMOR+ORDER+BY+info%3AQSS+DESC

PR #818, opened 2015-08-27
Fix VCF download feature

armish:
See #738 for more context.

When clicked on the download, we were getting this error logged:

```
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 59979)
Traceback (most recent call last):
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 321, in process_request
    self.finish_request(request, client_address)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 657, in __init__
    self.finish()
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/SocketServer.py", line 716, in finish
    self.wfile.close()
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 283, in close
    self.flush()
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 307, in flush
    self._sock.sendall(view[write_offset:write_offset+buffer_size])
error: [Errno 32] Broken pipe
----------------------------------------
```

which was stemming from a simple coding mistake.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/818)

### DIFF
--- a/cycledash/auth.py
+++ b/cycledash/auth.py
@@ -3,6 +3,7 @@ from flask import request, redirect, render_template
 from flask.ext.login import login_user, logout_user
 from sqlalchemy import exc
 import voluptuous
+import base64
 
 from cycledash import db, bcrypt, login_manager
 from cycledash.helpers import prepare_request_data, safely_redirect_to_next
@@ -107,3 +108,33 @@ def register():
             login_user(user)
             return redirect('/')
     return render_template('register.html', errors=errors)
+
+def check_login_from_key(api_key):
+    username, password = api_key.split(":")
+    return check_login(username, password)
+
+def load_user_from_request(request):
+    """Support for basic authorization."""
+    # first, try to login using the api_key url arg
+    api_key = request.args.get('api_key')
+    if api_key:
+        user = check_login_from_key(api_key)
+        if user:
+            return user
+
+    # next, try to login using Basic Auth
+    api_key = request.headers.get('Authorization')
+    if api_key:
+        api_key = api_key.replace('Basic ', '', 1)
+        try:
+            api_key = base64.b64decode(api_key)
+        except TypeError:
+            pass
+        user = check_login_from_key(api_key)
+        if user:
+            return user
+
+    # finally, return None if both methods did not login the user
+    return None
+
+login_manager.request_loader(load_user_from_request)
--- a/cycledash/auth.py
+++ b/cycledash/auth.py
@@ -115,13 +115,6 @@ def check_login_from_key(api_key):
 
 def load_user_from_request(request):
     """Support for basic authorization."""
-    # first, try to login using the api_key url arg
-    api_key = request.args.get('api_key')
-    if api_key:
-        user = check_login_from_key(api_key)
-        if user:
-            return user
-
     # next, try to login using Basic Auth
     api_key = request.headers.get('Authorization')
     if api_key:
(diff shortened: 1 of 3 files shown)


hammer:
is there a test you could add that would have caught this mistake?

armish:
yeah, it should do be doable using one of the test files we have. Let me add that and update the PR.

armish:
Add a test was much harder than I expected (due to some magic happening with authentication), but this is done now. On the bright side: we now have support for basic authentication for all pages within Cycledash.

cycledash/auth.py
```diff
@@ -107,3 +108,33 @@ def register():
...
             return redirect('/')
     return render_template('register.html', errors=errors)
+
+def check_login_from_key(api_key):
+    username, password = api_key.split(":")
+    return check_login(username, password)
+
+def load_user_from_request(request):
+    """Support for basic authorization."""
+    # first, try to login using the api_key url arg
```
ihodes: Mind dropping support for this? It's not the most secure (namely because URLs are stored in logs), and I'm not sure it buys us anything?

ihodes: Not a strong opinion, though, and if we need it in the future for some reason, then just go for it.

ihodes:
Looks great! One minor comment.

cycledash/auth.py
armish: @ihodes: without this one, my requests within `helpers.ResourceTest` were not being authorized and being forwarded to the `login` page automatically. I realized that all the `api` pages work fine with `ResourceTest`'s `get` and `post` requests, but none of our pages with `@login_required` decorator. 

I implemented this out of my ignorance about a better way; happy to hear your thoughts on this.

tests/python/test_download.py
```diff
@@ -0,0 +1,46 @@
...
+    def tearDown(self):
+        with tables(db.engine, 'projects', 'vcfs', 'genotypes') \
+                as (con, projects, runs, genotypes):
+            genotypes.delete().execute()
+            runs.delete().execute()
+            projects.delete().execute()
+
+    def test_download(self):
+        # match the URL with the default download button href
+        downUrl = ("/runs/" + str(self.run['id']) + "/download?query="
```
ihodes: This might be clearer if you write the query-string as JSON and then URL encode it in code--must easier to edit that way, too.

cycledash/auth.py
ihodes: Ah I just meant the url param version of this; keeping the BasicAuth version is :+1: 

Also, I'm confused a bit: I don't see an api_key urlparam in the download URL--how are you passing authorization though?

ihodes:
Looks good!

armish:
Thanks!

Issue #727, opened 2015-06-15
Timestamp is incorrect on "Recent Comments" view


jaclynperrone:
When I create a new comment, the timeline on the Project page says that it was created "4 hours" ago

![time-bug](https://cloud.githubusercontent.com/assets/5685044/8165589/a273d1a0-135f-11e5-9a10-18883c475975.png)

danvk:
@tavinathanson noooo! time zones strike again.

tavinathanson:
Ugh! :-1: 

If this is a regression, then that's on me for not introducing a test that would've caught it. I'll look into this!

tavinathanson:
@jaclynperrone I can't reproduce this locally or on `http://hammerlab-dev3.hpc.mssm.edu`; are you seeing this locally and/or elsewhere?

jaclynperrone:
Yup! I'm seeing it locally. I also noticed the bug during a demo on @danvk's local machine. So..maybe there is something to that

ihodes:
IIRC Tavi had to set the database time zone to UTC?

tavinathanson:
@ihodes Yeah, I'm thinking it's a Postgres settings thing, need to retrace my steps to remember exactly what I did.

tavinathanson:
@jaclynperrone Are you still hitting this issue?

jaclynperrone:
It looks fine on http://cycledash.demeter.hpc.mssm.edu/, but I am having the issue locally. I probably just need to configure something in my db?

tavinathanson:
I'm trying to figure out what you would need to configure, as we're discussing in #819.

Are you seeing the issue on the `cycledash` DB or the `cycledash-test` DB or both? What's your OS X timezone set to? And what's your `timezone` set to in `/usr/local/var/postgres/postgresql.conf`?

tavinathanson:
Also: if you go into your `cycledash` and `cycledash-test` DBs using `psql` or another client, do you see the following:

```
 created       | timestamp with time zone | not null default now()
 last_modified | timestamp with time zone | not null default now()
```

(Or do you get a `timestamp` or `timestamp without timezone` or something?)

jaclynperrone:
I get the error in both db's,
my OSX is set to EDT,
and `timezone` in /usr/local/var/postgres/postgresql.conf is US/Eastern

And when I run `\d user_comments` I also get:
```
created       | timestamp with time zone | not null default now()
last_modified | timestamp with time zone | not null default now()
```

[...]

PR #819, opened 2015-08-28
Suggest using UTC timezone for better compatibility with timestamps

armish:
I recently wiped my develop environment on Mac OS X 10.10 and with that came a new PostgreSQL installation. Everything was working fine but the `seltests` were failing due to different timestamps on the examine pages for various test runs. 

I was able to resolve this issue by switching my default timezone within `postgresql.conf` and thought that this was something specific to my setup; but, looks like issue #727 is also related to this and mentioning this within our `DEVELOP.md` helps clarifying this issue.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/819)

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -48,7 +48,11 @@ $EDITOR ENV.sh                     # Fill in values.
 
 ### 4. Set up the Database (Postgres)
 
-We use PostgreSQL as our datastore, with the schema described in `schema.sql`.
+We use PostgreSQL as our datastore, with the schema described in `schema.sql` and `UTC` timezone:
+
+```bash
+$EDITOR /usr/local/var/postgres/postgresql.conf  # timezone = 'UTC'
+```
 
 On OS X, setting up and running psql might look like this:
 


ihodes:
Hmm seltest tests should have all timestamps blanked out; which pages were timestamps showing up on?

ihodes:
Ah I see, this is from the Examine page timestamps?

armish:
yeah, the seltests for the Examine page were mysteriously failing me although everything looked the same... well, except the timestamps ;) This fixed the problem for me, not sure if there is a better way to fix this once and for all.

ihodes:
I just checked my database and my timezone is US/Eastern.... there must be something else doing on here?

armish:
hmm, that's interesting. Then setting the timezone is just a hack and it is possible that the problem I had was different than #727.

Do you think this new `timezone=True` has something to do with the issue I was having?

ihodes:
Yeah I'm really not sure what's going on here, @tavinathanson has a fair amount of experience with this, may be worth chatting. What is timezone=True?

tavinathanson:
Sorry for not chiming in earlier; didn't see this PR.

@ihodes, I'm confused about how your DB is Eastern and still working correctly! My local, and dev1, are both `UTC` (just double checked).

I'm going to set up CycleDash right now and see what happens when I mess with UTC vs. Eastern in the conf.

tavinathanson:
Hmm. I'm currently making comments, switching the Postgres setting between `UTC` and `US/Eastern`, making more comments, repeat. And all the comments are showing up correctly ("a few seconds ago", "2 minutes ago", etc.). So that doesn't seem to be the issue (or even necessary?).

Speaking of whether or not that's necessary, I appear to have said in #576: " (Though, that UTC setting in the conf file doesn't appear to be necessary, per: "it is assumed to be in the time zone indicated by the system's timezone parameter, and is converted to UTC using the offset for the timezone zone".)" (The inner quote is from http://www.postgresql.org/docs/9.1/static/datatype-datetime.html)

@armish, could your OS X timezone have gotten messed with?

armish:
thanks for checking this @tavinathanson!

This is what I have for my timezone settings:
screen shot 2015-09-01 at 3 18 36 pm

Maybe this is yet-another issue with my local machine and not a broad issue. Let's dismiss this PR until somebody else can replicate this problem.

With that, is it now safe to close #727 as well?

ihodes:
Re #727, unless we've fixed it, let's keep it open.

tavinathanson:
Re-opening because changing to `UTC` fixes the issue for both @jaclynperrone and @armish. @ihodes, what's different about your setup?!

ihodes:
Crazy! Maybe I do have the issue and have never noticed it? Either way, thanks for tracking it down.

tavinathanson:
Can you try creating a new comment to check?

ihodes:
It all appears to work for me! Maybe my postgres isn't using that config? http://link.isaachodes.io/image/301u080H0z1g

jaclynperrone:
@ihodes Is the timestamp also correct on the Runs page? It was wrong for me in the "Recent Comments" section.

tavinathanson:
So the two remaining mysteries are why non-UTC works for @ihodes and why it also works for me when I switch away from UTC. I guess that's not quite resolved.

ihodes:
(Off-thread we discovered that if I refreshed the page, the timestamp is out of date).

Hmm what do you mean "works"? It seems to not work given the refreshing showing an incorrect date?

tavinathanson:
Aha! So @ihodes's doesn't work after a refresh, when it actually pulls from the DB. Given that, I think this change should go in. Any objections?

ihodes:
Ahh got it :) Excellent. Thank for tracking this down! 

I'm still puzzled at how seltests were failing because of this (I haven't run them very recently, but I haven't had a timestamp issue with them in the recent past).

armish:
@ihodes: this is still a wild guess, but I think our schema change made those tests fail for me on examine pages. I didn't have this issue when were simply doing `< cycledash.sql`.

ihodes:
Weird—I will try running the seltest tests tomorrow to see! In the meantime
though, please merge on green, and sorry for the trouble!

armish:
no worries -- I have been having weird issues with my local setup lately, so I wasn't really sure whether this was because of my setup or not.

Merging this one in.

PR #820, opened 2015-08-28
Instead of forcing a default ENSEMBL release, infer it from the VCF via varcode

### LINKED ISSUE #816: Varcode worker: infer the correct ENSEMBL release from the VCF file if possible
We currently assume it is `Release 75` or [whatever is set in the configuration/env by default](https://github.com/hammerlab/cycledash/blob/master/workers/varcode_annotator.py#L52,L54). This is not always true. The worker should try to guess this from the file and if this is not available, then it should fall back to the default [one declared within config.py](https://github.com/hammerlab/cycledash/blob/master/config.py#L32). 

Related: #512 cc: @tavinathanson

armish:
Resolves #816. Note the schema change and the need to migrate.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/820)

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -39,6 +39,7 @@ Table('vcfs', metadata,
     Column('notes', String()),  # Any notes, params, etc the user might add. Ideally in JSON format.
     Column('uri', String()),  # URI of source file, if any
     Column('vcf_header', String()),  # Plaintext header of the VCF
+    Column('vcf_release', Integer),  # ENSEMBL compatible release for reference
     Column('extant_columns', String()),  # JSON list of non-null columns in the VCF
     Column('genotype_count', BigInteger),   # number of variants in this VCF
     UniqueConstraint('project_id', 'uri', name='vcfs_project_id_uri_key')
--- a/workers/shared.py
+++ b/workers/shared.py
@@ -100,17 +100,16 @@ def load_vcf(vcf_path):
     if config.ALLOW_LOCAL_VCFS and vcf_path.startswith('/tests/'):
         filepath = vcf_path[1:];
         text = open(filepath).read()
-        release = guess_ensembl_release(filepath)
     elif vcf_path.startswith('file://'):
         filepath = vcf_path[6:];
         text = open(filepath).read()
-        release = guess_ensembl_release(filepath)
     elif vcf_path.startswith('hdfs://'):
         return load_vcf(vcf_path[6:])
     else:
         text = get_contents_from_hdfs(vcf_path)
-        release = guess_ensembl_release(hdfs_to_local_path(vcf_path))
+        filepath = hdfs_to_local_path(vcf_path)
     header = '\n'.join(l for l in text.split('\n') if l.startswith('#'))
+    release = guess_ensembl_release(filepath)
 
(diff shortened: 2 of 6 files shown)


workers/shared.py
```diff
@@ -86,20 +87,32 @@ def hdfs_to_local_path(hdfs_path):
 
     return filename
 
+def guess_ensembl_release(filepath):
+    try:
+        release = varcode.load_vcf(filepath)[0].ensembl.release
```
ihodes: Does this parse the entire VCF to infer the release?

workers/varcode_annotator.py
```diff
@@ -42,16 +42,24 @@ def _annotate(vcf_id):
     if vcf_id == False:
         return  # An error must have occurred earlier.
 
+    engine = sqlalchemy.create_engine(DATABASE_URI)
+    # See if we have any guesses for a release version
+    with tables(engine, 'vcfs') as (con, vcfs):
+        metadata = sqlalchemy.MetaData(bind=con)
```
ihodes: You shouldn't have to construct or reflect the metadata here; `tables` should do that for you.

workers/shared.py
```diff
@@ -86,20 +87,32 @@ def hdfs_to_local_path(hdfs_path):
...
+    finally:
+        return release
 
 def load_vcf(vcf_path):
     """Return a vcf.Reader, header text for the given VCF."""
     if config.ALLOW_LOCAL_VCFS and vcf_path.startswith('/tests/'):
-        text = open(vcf_path[1:]).read()
+        filepath = vcf_path[1:];
+        text = open(filepath).read()
+        release = guess_ensembl_release(filepath)
```
ihodes: You can pull this `release = ...` out of the if/else block so you don't have to repeat it.

schema.py
```diff
@@ -39,6 +39,7 @@
     Column('notes', String()),  # Any notes, params, etc the user might add. Ideally in JSON format.
     Column('uri', String()),  # URI of source file, if any
     Column('vcf_header', String()),  # Plaintext header of the VCF
+    Column('vcf_release', Integer),  # ENSEMBL compatible release for reference
```
ihodes: You need to add a migration still! (docs in DEVELOPMENT.md)

workers/shared.py
armish: unfortunately yes. We have two other alternatives to this one:
1) Copy the logic within varcode over to Cycledash and pass the `vcf_header` field through that logic
2) Convince @iskandr to expose relevant guessing methods in https://github.com/hammerlab/varcode/blob/master/varcode/reference.py and use them directly

ihodes:
Awesome, thanks for handling this! Just a couple minor comments.

armish:
thanks for the comments, @ihodes. This should be good to go on green if you think the latest changes to the alembic script look reasonable.

ihodes:
LGTM :sparkles: :+1: :fireworks: :sparkles:

workers/shared.py
iskandr: @armish You can use `varcode.vcf.infer_genome_from_vcf` but then you're going to have to deal with the fact that not all VCFs are for humans :-)

PR #831, opened 2015-09-03
Associate comments with users and use username for author name

armish:
This PR deprecates #813 and adds support for proper associations between comments and their authors. With this change, we get rid of the `Author Name` field from the `CommentBox` and automatically assign user to a new comment: if the user logged in, then the comments belong to that user; if not, it is *anonymous*. If the user authentication is not enabled, all comments will be anonymous.

This PR requires a minor update to our DB schema and hence features an alembic migration script. During migration, all of our comments will lose their `author_name` information and therefore will become *anonymous*. 

Here are some screenshots to get a feeling of how this looks now:
screen shot 2015-09-03 at 1 38 19 pm

screen shot 2015-09-03 at 1 38 33 pm

screen shot 2015-09-03 at 1 38 45 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/831)


alembic/versions/1ab5aa1e3054_start_associating_comments_with_.py
```diff
@@ -0,0 +1,28 @@
+"""Start associating comments with registered users and use user information
```
ihodes: I'd frame these migration notes as 1) describing the change on the DB (e.g. "Add a user_id column to user_comments and drop author_name.") 2) then describe why (e.g. "This allows us to associate comments with users in the database.")

cycledash/api/comments.py
```diff
@@ -77,9 +77,11 @@ def get(self, run_id):
     @marshal_with(CommentFields)
     def post(self, run_id):
         """Create a comment."""
+        print current_user
```
ihodes: Remove the debug statement.

cycledash/static/js/examine/components/VCFTable.js
```diff
@@ -290,7 +294,7 @@ var VCFRecord = React.createClass({
     record: React.PropTypes.object.isRequired,
     columns: React.PropTypes.object.isRequired,
     isSelected: React.PropTypes.bool.isRequired,
-    handleSelectRecord: React.PropTypes.func.isRequired
+    handleSelectRecord: React.PropTypes.func.isRequired,
```
ihodes: NBD but don't need to add a comma here.

ihodes:
Awesome, thanks for doing this. I'm still looking through/commenting, but wanted to ask if the reason for the Anonymous user was for passing seltest tests (the only time I can think of that we have Anonymous users)?

armish:
Thanks for the comments @ihodes -- keep them coming :)

About the Anonymous user and the need for seltests: yes and no -- seltests are the only place where we set `LOGIN_DISABLED=true` and it helped uncover a bug: `current_user` cannot be serialized into JSON when the user is anonymous (see `examine.html`). I added that class just to make sure that it is serializable and the whole implementation behaves normally when the user anonymous (hence the `username` and `null` `id`). 

There might be other workarounds for this, but I thought that this might be good way for us to move forwards where we can customize the fields for Anonymous user and reduce the complexity of the code in other places.

cycledash/api/comments.py
```diff
@@ -174,8 +177,30 @@ def get_last_comments(n=5):
...
 
+def userify_comments(comments):
+    """Given comments with userIds, attaches the relevant user
+    info to the comments
+    """
+    for comment in comments:
+        userify_comment(comment)
+    return comments
+
+def userify_comment(comment):
```
ihodes: I'd call this something like `add_user_to_comment` to make it a little clearer that's going on (I see the parallels with `epochify_comment`, but it's doing something a bit different).

cycledash/auth.py
```diff
@@ -65,6 +65,24 @@ def is_anonymous(self):
     def get_id(self):
         return unicode(self['id'])
 
+class AnonymousUser(dict):
```
ihodes: You could inherit from User and from get_id (which, IIRC, should return unicode). If not that, make sure get_id returns unicode if it should (cf. above User).

cycledash/api/comments.py
```diff
@@ -174,8 +177,30 @@ def get_last_comments(n=5):
...
+    """Given a comment with userId, attaches the relevant user
+    info to the comment
+    """
+    if 'userId' in comment:
+        with tables(db.engine, 'users') as (con, users):
+            q = select(users.c).where(users.c.id == comment['userId'])
```
ihodes: `q = select([users.c.username, users.c.email, users.c.id])...`; generally better to whitelist than blacklist (which also lets you get rid of "desentisize_user" (if keeping it, spelling -> desensitize).

cycledash/api/comments.py
```diff
@@ -174,8 +177,30 @@ def get_last_comments(n=5):
...
+            user = q.execute().fetchone()
+            if user is not None:
```
ihodes: -> `if user:` is more "Pythonic" here.

cycledash/api/comments.py
```diff
@@ -174,8 +177,30 @@ def get_last_comments(n=5):
         q = select(cols.values()).order_by(
             desc(cols.created)).limit(n)
         comments = [camelcase_dict(dict(c)) for c in con.execute(q).fetchall()]
-    return epochify_comments(comments)
+    return userify_comments(epochify_comments(comments))
 
+def userify_comments(comments):
```
ihodes: "add_user_to_comments"

cycledash/api/comments.py
```diff
@@ -174,8 +177,30 @@ def get_last_comments(n=5):
...
+    """Given comments with userIds, attaches the relevant user
+    info to the comments
+    """
+    for comment in comments:
+        userify_comment(comment)
+    return comments
+
+def userify_comment(comment):
+    """Given a comment with userId, attaches the relevant user
+    info to the comment
```
ihodes: I'd note which columns it adds (email, username, id?).

tests/data.sql
```diff
@@ -109,12 +109,12 @@ SELECT pg_catalog.setval('task_states_id_seq', 2, true);
...
 --
 
-COPY user_comments (id, vcf_id, sample_name, contig, "position", reference, alternates, comment_text, author_name, created, last_modified) FROM stdin;
+COPY user_comments (id, vcf_id, sample_name, contig, "position", reference, alternates, comment_text, user_id, created, last_modified) FROM stdin;
 1	1	NORMAL	20	61795	G	T	This is a comment on the second row!	\N	2015-03-24 15:29:09.615967+00	2015-03-11 23:33:33.951753+00
 2	1	NORMAL	20	65900	G	A	*This* is a comment with [Markdown](http://daringfireball.net/projects/markdown/syntax).	\N	2015-03-24 15:29:09.615967+00	2015-03-11 23:34:18.19118+00
 3	1	TUMOR	20	61795	G	T	*This* is a comment on the first variant with [Markdown](http://daringfireball.net/projects/markdown/syntax).\n	\N	2015-03-24 15:29:09.615967+00	2015-03-12 15:30:20.570952+00
 4	1	NORMAL	20	61795	G	T	This is another comment on the second row!	\N	2015-03-24 17:40:44.984191+00	2015-03-24 17:40:44.984191+00
-6	1	TUMOR	20	61795	G	T	This is a comment without Markdown by Bob!	Bob	2015-03-24 17:46:40.962213+00	2015-03-24 17:46:40.962213+00
+6	1	TUMOR	20	61795	G	T	This is a comment without Markdown by Bob!	\N	2015-03-24 17:46:40.962213+00	2015-03-24 17:46:40.962213+00
```
ihodes: I'd add a fake user (or users) to this test and make this and other comments by those users, instead of a null user ID—that way we're testing this feature with the pdiff tests.

alembic/versions/1ab5aa1e3054_start_associating_comments_with_.py
```diff
@@ -0,0 +1,28 @@
...
+revision = '1ab5aa1e3054'
+down_revision = '2c00861f3ff1'
+branch_labels = None
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+
+def upgrade():
+    op.drop_column('user_comments', 'author_name')
```
ihodes: We're going to lose all existing author names, but that's okay :) I don't think that'll have a large impact for our current users. @arahuja

cycledash/api/comments.py
```diff
@@ -163,6 +165,7 @@ def _row_key(comment, table):
             comment['last_modified'] = to_epoch(comment['last_modified'])
             comment['created'] = to_epoch(comment['created'])
             comment = camelcase_dict(comment)
+            comment = userify_comment(comment)
```
ihodes: Update this function's docstring to note that you're adding a "user" field to comments.

ihodes:
This looks great; a few mostly minor comments. This really exposed you to the whole Cycledash stack now :) Thanks for handling this!

cycledash/static/js/examine/components/VCFTable.js
armish: I am surprised that lint didn't catch this.

cycledash/api/comments.py
armish: Argh!

armish:
Thanks @ihodes, this was really fun to work on. The latest changes address all your comments/suggestions.

ihodes:
Awesome, thanks--looks good to merge once you get it to go green!

Mind squashing those commits down a bit, first?

PR #837, opened 2015-09-14
Disable screenshot-based pdiff tests on Travis

armish:
We have been getting inconsistent screenshots from Sauce Labs across different runs/tests, which is causing us trouble with new feature branches. This will disable the part of tests run on Travis that compares the screenshots.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/837)

### DIFF
--- a/scripts/travis-test.sh
+++ b/scripts/travis-test.sh
@@ -12,9 +12,6 @@ echo 'Running tests in reverse...'
 echo 'Linting...'
 ./scripts/lint.sh
 
-echo 'Running seltests...'
-./scripts/travis-run-pdiff-tests.sh
-
 if [ $CI ]; then
   set +o errexit
   ./scripts/travis-coverage.sh


scripts/travis-test.sh
```diff
@@ -12,8 +12,10 @@ echo 'Running tests in reverse...'
 echo 'Linting...'
 ./scripts/lint.sh
 
-echo 'Running seltests...'
-./scripts/travis-run-pdiff-tests.sh
+# Seltests are disabled due to inconsistent screenshots
```
ihodes: I'd just kill these lines entirely; we can always get them back via source control!

ihodes:
Thank you! One little comment, hah :sparkles: :+1:

armish:
removed those lines completely as you suggested :D. Will merge on green.

Issue #840, opened 2015-09-18
Unable to submit VCF with Structural Variants


danvk:
I tried to submit [this VCF][1] which was produced by Manta. The genotype_extractor failed with this error:

```
Traceback (most recent call last):
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 437, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/cycledash/cycledash/workers/genotype_extractor.py", line 25, in extract
    return _extract(vcf_id)
  File "/home/cycledash/cycledash/workers/genotype_extractor.py", line 43, in _extract
    reader, header_text, release = load_vcf(vcf['uri'])
  File "/home/cycledash/cycledash/workers/shared.py", line 112, in load_vcf
    release = guess_ensembl_release(filepath)
  File "/home/cycledash/cycledash/workers/shared.py", line 96, in guess_ensembl_release
    return release
UnboundLocalError: local variable 'release' referenced before assignment
```

[1]: https://gist.github.com/danvk/1b909a87b0cac5e758b5

danvk:
Thanks for fixing this, @armish !

PR #841, opened 2015-09-18
Make genotype_extractor work even if varcode fails

armish:
Since #816, we were using varcode to infer the release of the VCF file; but when varcode fails #840 due to unexpected error, genotype extractor should still work and use the default release.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/841)

### DIFF
--- a/workers/shared.py
+++ b/workers/shared.py
@@ -92,6 +92,8 @@ def guess_ensembl_release(filepath):
         release = varcode.load_vcf(filepath)[0].ensembl.release
     except ValueError:  # no guesses from varcode, return default
         release = config.ENSEMBL_RELEASE
+    except Exception:  # varcode cannot handle this one, so go w/ default
+        release = config.ENSEMBL_RELEASE
     finally:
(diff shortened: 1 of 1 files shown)


workers/shared.py
```diff
@@ -92,6 +92,8 @@ def guess_ensembl_release(filepath):
         release = varcode.load_vcf(filepath)[0].ensembl.release
     except ValueError:  # no guesses from varcode, return default
         release = config.ENSEMBL_RELEASE
+    except Exception, error:  # varcode cannot handle this one, so go w/ default
```
ihodes: Probably don't need to capture the exception object (i.e. you can drop the `, error`) unless there's a reason to catch it?

ihodes:
LGTM on green % 1 comment

workers/shared.py
armish: my bad -- I first thought about logging it, then dropped that idea. Removed it.

Issue #828, opened 2015-09-02
Short VCFs have tooltips that don't fully appear


tavinathanson:
![image](https://cloud.githubusercontent.com/assets/322932/9641110/faf3ed40-5182-11e5-8812-06f2b7cbf8b5.png)

@jaclynperrone:
> @tavi That’s with a 1-row VCF?

@tavinathanson:
> @jaclyn: yeah http://cycledash.demeter.hpc.mssm.edu/runs/322/examine?query=1%3A61795-61796+ORDER+BY+contig%2C+position

@jaclynperrone:
> @tavi the table’s container’s  style is set to `overflow:auto`
> if you scroll, you see the rest of the tooltip

PR #850, opened 2015-10-02
Adds a min-height to table container

jaclynperrone:
This is to fix the tooltip bug seen in #828. I'm setting a `min-height` on the table container in order to show the overflowing tooltips. You'll notice that the min-height unit is in `vh`, which refers to "viewport height". This keeps the height flexible and relative to the height of the viewport, rather than using an arbitrary static `px` value. It's pretty awesome and widely [supported](http://caniuse.com/#feat=viewport-units). You can read more about it [here](http://stanhub.com/how-to-make-div-element-100-height-of-browser-window-using-css-only/).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/850)

### DIFF
--- a/cycledash/static/scss/examine.scss
+++ b/cycledash/static/scss/examine.scss
@@ -235,6 +235,7 @@ a.download-vcf {
 /*------[ VCF Table ]------*/
 .examine-table-container {
   overflow: auto;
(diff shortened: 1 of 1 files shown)


ihodes:
Neat fix!

PR #851, opened 2015-10-02
Add starring

### LINKED ISSUE #675: A user should be able to flag a variant
Right now comments are used as a way to both [1] flag a variant and [2] add a blurb about it. But these two actions serve two different purposes. This issue is about factoring out that desire to flag a variant for later inspection, and leaving the comment for...well, commenting (and reporting: #674)

This needs design, but it would act like a "favoriting" mechanism.

ihodes:
Fixes https://github.com/hammerlab/cycledash/issues/675

screen shot 2015-10-02 at 6 10 04 pm

WIP: fix tests & add tests for new feature.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/851)


jaclynperrone:
Oooo. Excited for this one!

danvk:
This mostly looks fine (thanks for implementing it!), but before we get too into the review, do you know why the fonts changed on the screenshots? e.g. in `examine_bad-query.png` the "Compare to..." text takes up much more of the box after this change.

ihodes:
Yeah I noticed that, too; the tests were (incorrectly) using the TypeKit engine for a few commits; this reverts it to the non-Typekit fonts. (We don't want people to need a Typekit account in order to run the screenshot tests).

danvk:
Reviewed 6 of 9 files at r1, 4 of 24 files at r2.
Review status: 10 of 31 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

<sup>**[cycledash/static/js/examine/RecordStore.js, line 274 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/851#-K096XlrXvJm8ZorzY1h)** ([raw file](https://github.com/hammerlab/cycledash/blob/1f148351fe7851558ede3a81e3470f65d24976bd/cycledash/static/js/examine/RecordStore.js#L274)):</sup>
`foo` and `realFoo` or `actualFoo` is a minor pet peeve for me. `realRecord` makes me question my understanding of what `record` was in the first place. It's more of an issue in languages where you don't have to declare the types of each variable.

How about names which indicate something more about what they are? e.g. `recordFromStore`. A comment would be helpful, too. I honestly don't know what's going on here.


---

<sup>**[cycledash/static/js/examine/RecordStore.js, line 333 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/851#-K097PHony7qS_WmKSii)** ([raw file](https://github.com/hammerlab/cycledash/blob/1f148351fe7851558ede3a81e3470f65d24976bd/cycledash/static/js/examine/RecordStore.js#L333)):</sup>
Also not sure how I feel about the naming here... just looking at the names, it seems clear what the difference between `starGenotype` and `starGenotypeAndNotify` should be. But that's not at all what the distinction between the functions is. Not sure what a better name would be, though. Maybe `setGenotypeStarAndNotify`?


---

<sup>**[tests/python/test_genotypes_api.py, line 48 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/851#-K098sINMbuPXSurZzgs)** ([raw file](https://github.com/hammerlab/cycledash/blob/1f148351fe7851558ede3a81e3470f65d24976bd/tests/python/test_genotypes_api.py#L48)):</sup>
how about a test where it fails, too? (e.g. because of an invalid contig)


---

<sup>**[workers/shared.py, line 157 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/851#-K098chz4GDUXbyXXox4)** ([raw file](https://github.com/hammerlab/cycledash/blob/1f148351fe7851558ede3a81e3470f65d24976bd/workers/shared.py#L157)):</sup>
This seems strange — if a column has TRUE/FALSE values set, isn't it "extant"?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/851)

danvk:
(LGTM, feel free to merge once you've addressed my comments & the tests pass.)


---

Review status: 10 of 31 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/851)

ihodes:
Review status: 7 of 31 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[workers/shared.py, line 157 \[r2\]](https://reviewable.io:443/reviews/hammerlab/cycledash/851#-K098chz4GDUXbyXXox4)** ([raw file](https://github.com/hammerlab/cycledash/blob/1f148351fe7851558ede3a81e3470f65d24976bd/workers/shared.py#L157)):</sup>
Yeah; this wasn't the right way to do it. What I explicitly want is a way to ignore certain columns. Starred is special in that I don't want it in the list of columns the VCFTable handles the display of (otherwise the it'd be displayed very differently); I want to handle its display myself. I added an EXCLUDED_FROM_EXTANT_COLUMNS variable to track columns I don't want to end up there.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/cycledash/851)

ihodes:
Thanks for the review! Let's see if Travis doesn't time out again (the tests look like they've been hanging on Travis… hopefully this got resolved overnight…)

PR #857, opened 2015-10-20
Handle Intragenic variants better

### LINKED ISSUE #855: Varcode annotator fails when there is an intragenic variant
```
Traceback (most recent call last):
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/celery/app/trace.py", line 437, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/cycledash/cycledash/workers/varcode_annotator.py", line 39, in annotate
    return _annotate(vcf_id)
  File "/home/cycledash/cycledash/workers/varcode_annotator.py", line 60, in _annotate
    annotations = get_varcode_annotations(genotypes, vcf_id, release)
  File "/home/cycledash/cycledash/workers/varcode_annotator.py", line 135, in get_varcode_annotations
    notation = best_effect.short_description
  File "/home/cycledash/cycledash/venv/lib/python2.7/site-packages/varcode/effects.py", line 45, in short_description
    self.__class__.__name__,))
ValueError: Property short_description not implemented for Intragenic
```

armish:
This fixes #855 and adds a new intragenic variant to the test file to prevent possible regression in the future.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/857)


armish:
cc @iskandr

workers/varcode_annotator.py
```diff
@@ -132,7 +132,10 @@ def get_varcode_annotations(genotypes, vcf_id, ensembl_release_num):
         best_effect = variant.effects().top_priority_effect()
         gene_name = best_effect.gene_name
         transcript = best_effect.transcript_id
-        notation = best_effect.short_description
+        if best_effect.__class__.__name__ == "Intragenic":
```
ihodes: I'd use `isinstance(best_effect, Intragenic)` (if that works), instead. Does Intragenic have no short_description? Is that something that should be fixed in varcode? (If so, file an issue?)

ihodes:
LGTM, thank you for the fix! One comment.

workers/varcode_annotator.py
armish: thanks @ihodes! I did it this way because there are a few other classes of variants that implement Intragenic but have their short_description implemented. AFAIK, this is the only class that doesn't have that field.

armish: confirmed on Slack: it is a varcode bug.

PR #858, opened 2015-10-21
Update to React 0.14

danvk:
Highlights:
- `React` → `ReactDOM`
- Drop many `.getDOMNode()` calls
- `React.addons.classSet` → `require('classnames')`
- `React.addons.TestUtils` → `require('react-addons-test-utils')`
- `ReactClass.type` → `ReactClass`
- Updated to `pileup@0.5.1`

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/cycledash/858)

### DIFF
--- a/cycledash/static/js/runs/components/RunsPage.js
+++ b/cycledash/static/js/runs/components/RunsPage.js
@@ -279,7 +279,7 @@ var RunRow = React.createClass({
     handleClick: React.PropTypes.func.isRequired
   },
   handleClick: function(evt) {
-    if (evt.target == this.refs.link.getDOMNode()) return;
+    if (evt.target == this.refs.link) return;
     this.props.handleClick();
   },
   render: function() {
--- a/cycledash/static/js/comments/comments.js
+++ b/cycledash/static/js/comments/comments.js
@@ -1,8 +1,9 @@
 'use strict';
 var React = require('react'),
+    ReactDOM = require('react-dom'),
     CommentsPage = require('./components/comment').CommentsPage;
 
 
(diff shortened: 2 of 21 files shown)


cycledash/static/js/runs/components/RunsPage.js
```diff
@@ -125,13 +126,11 @@ var RunsPage = React.createClass({
...
           </div>
           <div className="projects-table">
-          <div className='filter-runs'>
-            <label>Now Viewing</label>
-            <select className='select-project-filter' value={this.state.projectFilter}
-                    onChange={this.handleProjectFilter}>
-              {projectOptions}
-            </select>
-          </div>
+            <div className='filter-runs'>
```
ihodes: Doesn't seem like this file should be changed?

ihodes:
Thanks for debugging this. Is it worth doing this in two steps? Seems like the 0.13 change is just a massive regression in terms of functionality, would be better to do 0.12→0.14. Are you planning on upgrading this to 0.14 right away?

cycledash/static/js/runs/components/RunsPage.js
danvk: Yeah, just an indentation fix. Reverted.

danvk:
Yeah, good point. I'll update to 0.14 and push another commit to this branch.

ihodes:
Cool, sounds good, thanks!

danvk:
This depends on the next version of pileup (see https://github.com/hammerlab/pileup.js/issues/246), which I hope to release tomorrow.

ihodes:
LGTM, thanks for doing this!

